PR #1, opened 2023-05-29
[13.0][ADD] sign_oca: New module

victoralmau:
New module

@Tecnativa TT41745


rvalyi:
Hello sorry if that's a dump question: but do you plan to develop a module that will enable a customer to sign some pdf in the portal or only in the Odoo backoffice interface as it seems the case today?

pedrobaeza:
Not for now, Raphaël. It should be straight-forward though looking to `sale` module and their sign option.

sign_oca/models/sign_request.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class SignRequest(models.Model):
+    _name = "sign.request"
```
etobella, line 9: Calling it `sign.request` might be a problem, as odoo already has this model name for an enterprise module

pedrobaeza: Well, we designed this even without looking at enterprise. I don't think both modules are going to be installed together, but no special problem to rename it. Which name has you used?

etobella: In the module that I made on 14, there are two kind of records: sign.oca.request and sign.oca.request.signer

The first one assigns something to be signed. The second one is created for each signer.

pedrobaeza: Well, a name a bit weird. I think it's the first time we include `oca` in the model name, but if you still think we should avoid the name conflict, go ahead. On other modules like `account_asset_management`, we overlapped model names during years.

victoralmau:
Superseed by https://github.com/OCA/sign/pull/8

PR #6, opened 2023-08-28
[14.0][ADD] Add Lex Persona basic modules

cvinh:
TODO :
- [ ] Fix mixing models between connector and base (thanks to @etobella )
- [ ] Check how to have an abstract model or a connector as noticed here : https://github.com/OCA/sign/pull/6#issuecomment-1696874880
- [ ] Improve documentation with videos and clear explanations (I can manage that)
- [ ] Develop unit tests (help needed)
- [ ] We should be able to place the signature fields in odoo's wizard before sending the documents to the API (@etobella can you help please ?)

### DIFF
--- a/lexpersona_base/__init__.py
+++ b/lexpersona_base/__init__.py
@@ -1,0 +1,4 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import models
+from . import wizard
--- a/lexpersona_base/__manifest__.py
+++ b/lexpersona_base/__manifest__.py
@@ -1,0 +1,33 @@
+# Copyright 2022 Cyril VINH-TUNG - INVITU
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Lex Persona Base",
+    "version": "14.0.1.0.0",
+    "summary": "Base to Lex Persona API",
+    "sequence": 30,
+    "category": "Customization",
+    "author": "INVITU, " "Odoo Community Association (OCA)",
+    "maintainers": ["invitu"],
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/sign",
+    "images": [],
+    "depends": [
+        "base",
+        "mail",
+        "partner_firstname",
+        "partner_contact_nationality",
+        "lexpersona_connector",
+    ],
+    "data": [
+        "security/ir.model.access.csv",
+        "data/lexpersona_base_data.xml",
+        "data/cron.xml",
+        "views/ir_attachment_views.xml",
+        "wizard/wizard_create_workflow_views.xml",
+    ],
+    "demo": [],
+    "qweb": [],
+    "installable": True,
+    "application": False,
+    "auto_install": False,
+}
--- a/lexpersona_base/data/cron.xml
+++ b/lexpersona_base/data/cron.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="1">
+        <record id="ir_cron_ir_attachment_lexpersona" model="ir.cron">
+            <field name="name">Get signed documents from Lex Persona</field>
+            <field name="model_id" ref="model_ir_attachment" />
+            <field name="state">code</field>
+            <field name="code">model._retrieve_documents()</field>
+            <field name="interval_number">10</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field name="doall" eval="False" />
+            <field name="active" eval="True" />
+        </record>
+    </data>
+</odoo>
--- a/lexpersona_base/data/lexpersona_base_data.xml
+++ b/lexpersona_base/data/lexpersona_base_data.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="1">
+        <template id="sign_notification_template">
+            <div>
+                <p>Hello</p>
+                <p
+                >This is an automated message to inform you that the documents have been signed in <strong
+                        t-esc="record.display_name"
+                    /></p>
+                <br />
+                <br />
+                <p t-if="note" t-esc="note" />
+            </div>
+        </template>
+    </data>
+</odoo>
--- a/lexpersona_base/models/__init__.py
+++ b/lexpersona_base/models/__init__.py
@@ -1,0 +1,3 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import ir_attachment
--- a/lexpersona_base/models/ir_attachment.py
+++ b/lexpersona_base/models/ir_attachment.py
@@ -1,0 +1,67 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+import base64
+
+from odoo import _, api, fields, models
+
+
+class IrAttachment(models.Model):
+    _inherit = ["ir.attachment"]
+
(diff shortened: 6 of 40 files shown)


cvinh:
@etobella

etobella:
As we are unable to test it (At least, I don't have a lex persona account). Can you send a video about it?

Also, I don't understand why you splitted in two modules `lexpersona_base` and `lexpersona_connector`. Models are even mixed between both modules (you are calling models defined in the child module)

BTW, lexpersona is prepared to work for other countries? Website is only in French and is talking about France, I don't know if it is prepared for other countries (even in the EU)

Just to finish my questions, I don't understand the models `wizard.signature.type`and `wizard.signature.profile` shouldn't that be inserted on the specific models. They are created on the fly just to fill this data when retrieving the files, isn't it?

cvinh:
> As we are unable to test it (At least, I don't have a lex persona account). Can you send a video about it?

@etobella I can create a free account with enough credit for you to test it. I also need your help to move the signature field on the pdfs. Are you part of it ?

cvinh:
> Also, I don't understand why you splitted in two modules `lexpersona_base` and `lexpersona_connector`. Models are even mixed between both modules (you are calling models defined in the child module)

Because I expect lexpersona_connector to be used outside the backend (from an external App connected to odoo's rpc or from the portal without using the backend)
lexpersona_base is the basis of the backend

cvinh:
> BTW, lexpersona is prepared to work for other countries? Website is only in French and is talking about France, I don't know if it is prepared for other countries (even in the EU)

You are right, they focus on the French market I can ask their CEO why they did not translate their website... BTW I can push Fenua Sign (powered by Lex Persona) to other countries easily at least to EU and South Pacific Region (I am Fenua Sign's CEO)

cvinh:
> Just to finish my questions, I don't understand the models `wizard.signature.type`and `wizard.signature.profile` shouldn't that be inserted on the specific models. They are created on the fly just to fill this data when retrieving the files, isn't it?

These models are only used in the wizard, I think we don't need them in the models.
They are needed when we send the files to the API to tell which type of signature (simple, advanced, qualified) and which profile to launch (available options are visible signature or not, if the documents need to be read to the end or not, some documents can be attached to the signature workflow, just to be read not to be signed..)
You might understand everything with a real use.

etobella:
> > Also, I don't understand why you splitted in two modules `lexpersona_base` and `lexpersona_connector`. Models are even mixed between both modules (you are calling models defined in the child module)
> 
> Because I expect lexpersona_connector to be used outside the backend (from an external App connected to odoo's rpc or from the portal without using the backend) lexpersona_base is the basis of the backend

But you are mixing things, defining models in the child module that will be defined in the parent module. Also, but the kind of model  you are defining for the api model, I would recommend an abstract model or a connector.

My recommendation would be to keep it all in a single module (probably, no other module will use the frontend module to define a new api in top of it)

pedrobaeza:
The important question is if this can be used outside of France. If not, this should go on OCA/l10n-france instead of here.

etobella:
> > As we are unable to test it (At least, I don't have a lex persona account). Can you send a video about it?
> 
> @etobella I can create a free account with enough credit for you to test it. I also need your help to move the signature field on the pdfs. Are you part of it ?

I would recommend to use sign_oca in order to define where the fields are stored. The JS part only needs a model with the following fields and methods:
    
- datas: The PDF binary file
- get_info: method that returns all the current data to configure it
- delete_item: method that adds a item item
- set_item_data: method to modify a signature item
- unlink_item: method to unlink a signature item

cvinh:
> I would recommend to use sign_oca in order to define where the fields are stored. The JS part only needs a model with the following fields and methods:

Humm thanks for your point of you ❤️ but you might consider that Lex Persona allows to sign several documents and having other documents to be read but not signed in the same process. It's a real workflow manager, much more than a simple signature tool.
For example, a Lex Persona workflow can be made as follow : 
- several documents to be signed
- several documents to be approved only (not signed but part of the workflow)
- users can approve the documents
- users can sign the documents
- you can organize the steps of the workflow in any way (approval by someone first, then signature by anotherperson, then approval again by another person, then signature by another person, then approved again by the first person... and it can go to infinite)

You might say that these steps can be made in Odoo but if they are done in Lex Persona, the workflow is entirely legally proven and the proof of the workflow is kept 10 years by Lex Persona.
That's one of the reasons why we choose Lex Persona for our country among other solutions. We think it's the most powerful and flexible solution out of them, and we were happy that they were french 😆 ... no offense, we like catalans and bascos too ❤️ 🤣 💋 

@pedrobaeza Lex Persona covers EiDAS laws so it can be considered at least as European Solution (even if they didn't translate their website 👎). BTW, you might consider Fenua Sign which is definitely opened to other countries (it's a Lex Persona tenant)... website will be translated in english and spanish soon 🚀

cvinh:
> But you are mixing things, defining models in the child module that will be defined in the parent module.

Ah... then it's a mistake, I will check it, thanks !

etobella:
> > I would recommend to use sign_oca in order to define where the fields are stored. The JS part only needs a model with the following fields and methods:
> 
> Humm thanks for your point of you ❤️ but you might consider that Lex Persona allows to sign several documents and having other documents to be read but not signed in the same process. It's a real workflow manager, much more than a simple signature tool. For example, a Lex Persona workflow can be made as follow :
> 
> * several documents to be signed
> * several documents to be approved only (not signed but part of the workflow)
> * users can approve the documents
> * users can sign the documents
> * you can organize the steps of the workflow in any way (approval by someone first, then signature by anotherperson, then approval again by another person, then signature by another person, then approved again by the first person... and it can go to infinite)
> 

I understand that. But you were asking about the possibility to edit the location of the signature on a file, isn't it? The configuration of this can be done with the sign_oca JS code and do the signature process in lexpersona, isn't it?

I would not recommend to replicate all the signature info there. BTW, I am reviewing how to make the signature process from sign_oca as an Advanced Signature following eIDAS design. It should be possible and all information would be integrated on Odoo. We should remember that an advanced signature is legally binding (https://en.wikipedia.org/wiki/Advanced_electronic_signature). Some discussion might arise of how to make it, but with a meetings we would probably had enough ideas and quorum  (Maybe in Liége?).


> You might say that these steps can be made in Odoo but if they are done in Lex Persona, the workflow is entirely legally proven and the proof of the workflow is kept 10 years by Lex Persona. That's one of the reasons why we choose Lex Persona for our country among other solutions. We think it's the most powerful and flexible solution out of them, and we were happy that they were french 😆 ... no offense, we like catalans and bascos too ❤️ 🤣 💋
> 
> @pedrobaeza Lex Persona covers EiDAS laws so it can be considered at least as European Solution (even if they didn't translate their website 👎). BTW, you might consider Fenua Sign which is definitely opened to other countries (it's a Lex Persona tenant)... website will be translated in english and spanish soon 🚀

cvinh:
> Some discussion might arise of how to make it, but with a meetings we would probably had enough ideas and quorum (Maybe in Liége?).

I am considering going to Liege, but it's 24 hours flight... before that we can do a jitsi meet, I will contact you in private
...and I will update screenshots and videos soon
Stay tuned

cvinh:
> But you are mixing things, defining models in the child module that will be defined in the parent module.

@etobella can you please 🙏  point out where I mixed models in child and parent ? I cannot find it...

etobella:
Sorry,  my mistake. You are not mixing things, but making lexpersona_base depend on lexpersona_connector is really confusing :cry: 

If you are calling it base, it should be the first module IMO. We should rename it to lexpersona_interface and the connector to lexpersona_base, WDYT? At the end, the interface requires the Connector. So, for me, it is the base

PR #8, opened 2023-09-28
[14.0][IMP] sign_oca: Systray management + user rules + wizard to create requests

victoralmau:
Related to: https://github.com/OCA/sign/pull/1 (Some initial approaches or ideas are introduced)

Changes done:
- [x] Hide Sign menu for users who do not have permissions.
- [x] Add the possibility to link a record to a request (`record_ref` field).
- [x] Add rules for models + new Administrator group to configure / delete some data.
- [x]  Basic user will have read permissions to be able to sign.
- [x] Add systray compatibility (linked to signer records). Only for requests with `record_ref` set.
- [x] Add sign button to `sign.oca.request` + `sign.oca.request.signer` to allow any user access to sign page.
- [x] Add a model (optional) to the templates (to be explained).
- [x] Improve roles to set partner type (empty, default or expression)
- [x] Add a wizard to create requests based on a template and the selected records.

@Tecnativa TT41745


sign_oca/tests/test_sign_portal.py
```diff
@@ -47,6 +47,7 @@ def setUp(self):
         )
 
     def test_portal(self):
+        self.authenticate("portal", "portal")
```
etobella, line 50: Why do you need to authenticate? Previously this was not needed :thinking:

victoralmau: Without this change, when I run the test locally I get the error:

`requests.exceptions.HTTPError: 404 Client Error: NOT FOUND for url: http://127.0.0.1:8069/sign_oca/document/8/xxxxxxx`

Other tests of other modules that use `self.url_open()` also set `self.authenticate()` before.

pedrobaeza:
Please avoid generic descriptions like "Misc. changes". In this case, the goals are very clear: "Direct signing without PDF" and "Systray management".

victoralmau:
> Please avoid generic descriptions like "Misc. changes". In this case, the goals are very clear: "Direct signing without PDF" and "Systray management".

You are right, I will have to think of a better title for all these changes.

Finally there is another approach (create requests automatically or manually according to a template) to not sign without the PDF and deform the actual behavior.

victoralmau:
Ping @pedrobaeza

elvise:
@pedrobaeza gentle reminder :)

pedrobaeza:
@elvise you can perform a review as well. I will as soon as possible.

sign_oca/static/src/elements/systray.xml
```diff
@@ -0,0 +1,81 @@
...
+        </t>
+    </t>
+    <t t-name="sign.oca.RequestMenu">
+        <li class="o_mail_systray_item">
+            <a
+                class="dropdown-toggle o-no-caret"
+                data-toggle="dropdown"
+                data-display="static"
+                aria-expanded="false"
+                title="Requests"
```
bosd:
```suggestion
                title="Sign Requests"
```

sign_oca/models/sign_oca_template.py
```diff
@@ -106,3 +155,17 @@ def get_info(self):
             "height": self.height,
             "placeholder": self.placeholder,
         }
+
+    def _get_full_info(self):
+        """Método usado en los wizards en los requests que se crean."""
```
bosd: Can you please make the comment in English

sign_oca/models/sign_oca_request.py
```diff
@@ -295,6 +350,16 @@ def get_info(self, access_token=False):
             },
         }
 
+    def action_sign_url(self):
+        self.ensure_one()
+        if not self.is_allow_signature:
+            raise ValidationError(_("You can't sign this"))
```
bosd:
```suggestion
            raise ValidationError(_("You are not allowed to sign this document."))
```

sign_oca/security/security.xml
```diff
@@ -14,9 +14,113 @@
...
+        >[('partner_id', 'child_of', [user.partner_id.commercial_partner_id.id])]</field>
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+    </record>
+    <!-- User Record Rules -->
+    <record id="sign_oca_template_rule_user_read" model="ir.rule">
+        <field name="name">Sign Oca Template user: read</field>
```
bosd: As a question/remark in general. Do we need all the fields to be OCA branded?
Does not add value to the end user.
It is very unlikely someone will install odoo sa and oca's sign module.
Adn dev's will look at xmlid anyway.

bosd (commented):
Thanks for your work!
Did'nt do complete functional test yet.
This is code review and some UI remarks.

sign_oca/security/security.xml
victoralmau: Removed OCA from the name.

pedrobaeza:
@victoralmau all the new options require to expand the README, also for people to test the PR. One question: why not showing in the systray icon all the sign requests, no matter if linked to a record or not?

victoralmau:
> @victoralmau all the new options require to expand the README, also for people to test the PR. One question: why not showing in the systray icon all the sign requests, no matter if linked to a record or not?

You are right, I will add the new options to the README to clarify it.
In the systray only requests linked to models are shown because as with `base_tier_validation` or activities, they need to have a model linked so they can be shown in a grouped way.

pedrobaeza:
> In the systray only requests linked to models are shown because as with base_tier_validation or activities, they need to have a model linked so they can be shown in a grouped way.

You can group them in an "Undefined" section.

victoralmau:
Changes done.

elvise:
> @elvise you can perform a review as well. I will as soon as possible.

@pedrobaeza what do you think about adding a smart button in the partner form to link all signatures?

bosd (approved):
LGTM now

sign_oca/models/res_users.py
```diff
@@ -0,0 +1,50 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models, modules
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def sign_oca_request_user_count(self):
```
bosd, line 11: We should re-use odoo's activities systray, to attack serious perf issues.

bosd (changes requested):
Coming back to this one.
Adding yet another systray is bad for performance.

Odoo had a lot of issues/pr's regards of the performance of the systray.
There is as example an open PR  136851 on the performance of the systray.
The argument there is that code is executed on every page load.
Resulting that the systray is slowing down the loading of the homepage.


As having one systray for activities (odoo core) is impactfull for performance.
We keep adding.
The systray code here looks inspired from the oca `base_tier validation`.
(which I'm working on to refactor for perfmance)
So we'll get..
activities systray (odoo core)
oca base_tier_validation systray
oca sign systray
oca .... systray module (future)

All adding up to page load speed.


If we add a systry icon, in case for performance and maintainability we should re-use the odoo activities systray.
Odoo core has seen a lot of refactors already trying to battle there performance issues.
Judging by the code on the oca modules. It is hard for us to keep up.
When we re-use their mechanism on oca modules. We also benefit from their perf improvements.

victoralmau:
Sorry, but I don't feel the same way about systray. The idea is to have few records linked to each user, the PR you indicate in which 10000 activities are mentioned for one user, I think it is not a real use case at all.

BernatPForgeFlow (approved):
Functional testing looks good!

victoralmau:
Finally, what do you think @pedrobaeza and @etobella about this approach?

sign_oca/views/sign_oca_request.xml
```diff
@@ -32,11 +39,11 @@
                 <sheet>
                     <div class="oe_button_box" name="button_box">
                         <button
-                            name="preview"
-                            type="object"
                             class="oe_stat_button"
                             icon="fa-pencil"
                             states="sent,signed"
+                            type="object"
```
etobella: Why? Info is already on the notebook of the record?

We can add the button to sign there and we avoid to do too many click. From the button, we could go to the documents to sign. WDYT?

etobella (commented):
Small comment, Also, I did not receive a bus notification with the document to sign (it showed nothing until I pressed it and then it showed that there was a document to sign)

sign_oca/views/sign_oca_request.xml
victoralmau: Sorry, it is already changed (previously there was another approach perhaps more confusing). 
- The existing smart-button is kept (groups are added to avoid error for "_basic_" users without sign permissions), 
- A Sign button (`sign.oca.request` form) is added to be able to sign (for example a "basic" user without sign permissions).

victoralmau:
> Small comment, Also, I did not receive a bus notification with the document to sign (it showed nothing until I pressed it and then it showed that there was a document to sign)

Ok, I see what you mean.
I think it's something that can be added to ROADMAP, reloading would be enough, although I will investigate how to get it.

pedrobaeza:
@etobella do you know how to send that bus notification?

etobella:
Yes, but not too easy to write the solution in a comment :thinking: 

we could proceed with this and fix it in Tenerife, WDYT?

victoralmau:
Ok, no problem.

Also comment that exactly this same problem is currently happening in other modules (`base_tier_validation` for example).

bosd:
> fix it in Tenerife

Wauw, you guys go on holidays together? 😄 🏖️

pedrobaeza:
Hehe, half of it: https://www.aeodoo.org/en/event/xvii-jornadas-odoo-116/page/introduccion-xvii-jornadas-odoo-copy

You have to come if you want to enjoy an incomparable landscape with some work.

pedrobaeza (approved):
/ocabot merge major

victoralmau:
Can you try again?

PR #14, opened 2023-12-01
[16.0][ADD] sign_biometric_oca: Proof of Concept

etobella:
Theoretically, this module allows us to store position, pressure and time of the signature, making it biometric.

I still need to compare with  https://www.iso.org/standard/77910.html in order to check if something is missing.

Theoretically, we need:

- Position
- Pressure
- Speed
- Acceleration

Speed and Acceleration might be computed using position and time, so I assume everything is correcly computed :smile:


gdgellatly:
If I recall correctly speed is the first order rate of change of distance, distance is the first order rate of change of position, and acceleration is the first order rate of change of speed, all over time, so should all be able to be computed but it will probably need multiple positions and speeds measured. But I imagine some 3rd party lib does this already.

etobella:
Exactly, we could compute it on the fly if we want to :smile:

lubusax:
How could pressure be measured?

Android has a built-in function to measure pressure:
https://developer.android.com/reference/android/view/MotionEvent.html#getPressure(int)

getPressure() is a method in the MotionEvent class in Android that returns the pressure of the touch event. The pressure value is a float that ranges from 0.0 to 1.0, where 1.0 represents the maximum pressure that the device can detect.

etobella:
> How could pressure be measured?
> 
> Android has a built-in function to measure pressure: https://developer.android.com/reference/android/view/MotionEvent.html#getPressure(int)
> 
> getPressure() is a method in the MotionEvent class in Android that returns the pressure of the touch event. The pressure value is a float that ranges from 0.0 to 1.0, where 1.0 represents the maximum pressure that the device can detect.

I think pressure is not required, but recommended. I need to check on a mobile how it was stored, but the JS worked perfectly :smile:

etobella:
Just be aware that this module should not be used on production right now, as you will be collecting biometric data and storing it in a raw format inside Odoo!!!


I am still working on it, but I hope I can send a final version soon :smile:

etobella:
Biometric data should be secure with this new change. Still reviewing it, but that should be enough :smile:

etobella:
In order to make it work, you might need a certificate.

To generate it, just use the following code:

```bash
openssl ecparam -name secp256k1 -genkey -noout -out ec-secp256k1-priv-key.pem
openssl ec -in ec-secp256k1-priv-key.pem -pubout > ec-secp256k1-pub-key.pem
```

If you want the value to paste into odoo, you can use:

```bash
openssl ec -in ec-secp256k1-priv-key.pem -pubout -outform DER  | base64 -w 0
```

pedrobaeza:
There are conflicts.

pedrobaeza (commented):
Why is this adding also things about PDF?

etobella:
Because it creates a completly different widget for images. The system needs to know how this data is displayed inside a PDF. The original widget is not right, because signing writing your name with your keyboard is, obviously, not a biometric.

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

PR #18, opened 2024-02-27
[14.0][FIX] sign_oca: Some fixes

PauBForgeFlow:
1. Add _Sign_ button in request form (it was only in tree view).
2. Fix configure template action when it is created from a request.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -99,7 +99,9 @@ class SignOcaRequest(models.Model):
     def _compute_next_item_id(self):
         for record in self:
             record.next_item_id = (
-                record.signatory_data and max(record.signatory_data.keys()) or 0
+                record.signatory_data
+                and max([int(key) for key in record.signatory_data.keys()])
+                or 0
             ) + 1
 
     @api.depends_context("uid")
--- a/sign_oca/views/sign_oca_request.xml
+++ b/sign_oca/views/sign_oca_request.xml
@@ -10,6 +10,14 @@
             <form create="0">
                 <header>
                     <button
+                        name="sign"
+                        class="btn btn-primary"
+                        type="object"
+                        string="Sign"
+                        attrs="{'invisible': [('to_sign', '=', False)]}"
+                    />
+                    <field name="to_sign" invisible="1" />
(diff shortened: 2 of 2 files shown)


BernatPForgeFlow (approved):

sign_oca/models/sign_oca_request.py
```diff
@@ -73,7 +73,7 @@ class SignOcaRequest(models.Model):
     def _compute_next_item_id(self):
         for record in self:
             record.next_item_id = (
-                record.signatory_data and max(record.signatory_data.keys()) or 0
+                record.signatory_data and int(max(record.signatory_data.keys())) or 0
```
etobella: Reviewing this is probably a bad decision, we should need to do:
`max([int(key) for key in record.signatory_data.keys()])`

For example, if you do max(['9', '10']) the result is '9'

sign_oca/models/sign_oca_request.py
```diff
@@ -102,7 +102,7 @@ def _ensure_draft(self):
 
     def configure(self):
         self._ensure_draft()
-        self._set_action_log("configure")
+        self._set_action_log("view")
```
etobella: Why do you change this?

etobella (changes requested):

sign_oca/models/sign_oca_request.py
PauBForgeFlow: There is not an action named "configure". Another change could be to create it.

victoralmau: IMO the correction should be https://github.com/OCA/sign/pull/8/commits/e0591479ae850c5b7b02cb7d563fb5f49d2ff2c3

etobella: I do agree with @victoralmau

PauBForgeFlow: Deleted. This will be fixed with @victoralmau commit.

pedrobaeza:
Can you rebase?

pedrobaeza (approved):
/ocabot merge minor

pedrobaeza:
@etobella I have launched the merge command as your requests were attended.

PR #19, opened 2024-02-28
[14.0][IMP] sign_oca: Link from partner to signer

BernatPForgeFlow:
Depends on: https://github.com/OCA/sign/pull/8


victoralmau (commented):
Code and functional review OK.

Won't really work until related is merged.

victoralmau (approved):

sign_oca/views/res_partner_views.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <div name="button_box" position="inside">
+                <button
+                    class="oe_stat_button"
+                    type="object"
+                    name="action_show_signer_ids"
+                    attrs="{'invisible': [('signer_count', '=', 0)]}"
+                    icon="fa-pencil"
+                    groups="sign_oca.sign_oca_group_user"
```
pedrobaeza: For this version, the groups should be put at view level, not element level, or you will have an error.

BernatPForgeFlow: I'm not sure if I understand what you mean. In version 14, groups can be added to buttons to not be shown. I have tested it and it works fine.

pedrobaeza: As said, if you enter in this view without having the permissions, it will give an access error. Please put the group at view level, not element level. Something like:

```xml
    <record id="view_partner_form" model="ir.ui.view">
        <field name="name">res.partner.form (in sign_oca)</field>
        <field name="model">res.partner</field>
        <field name="inherit_id" ref="base.view_partner_form" />
        <field name="groups_id" eval="[(4, ref('...')]" />
        <field name="arch" type="xml">
            <div name="button_box" position="inside">
```

victoralmau (approved):
Code and functional review OK.

I think sometimes there will be confusion of `sign.request` linked to a partner vs `sign.request.signer`, but it is "_normal_".

sign_oca/views/res_partner_views.xml
BernatPForgeFlow: I do not agree that it would give access error. However, I change it since it seems right to me anyway.

pedrobaeza (approved):
/ocabot merge minor

pedrobaeza:
Oh, wait, there are conflicts, so they should be resolved before merging. @BernatPForgeFlow would you?

sign_oca/views/res_partner_views.xml
pedrobaeza: If there are no access errors, it's because this model is readable for all users, but it's a good practice to do it this way because:

- It avoids needing to access the model without need.
- It avoids access errors when the model is not readable.

BernatPForgeFlow:
> Oh, wait, there are conflicts, so they should be resolved before merging. @BernatPForgeFlow would you?

@pedrobaeza  Can you merge now?

pedrobaeza:
/ocabot merge minor

@victoralmau to be cherry-pick in the v15 migration

PR #20, opened 2024-03-05
[14.0][IMP] sign_oca: Add Sign from template action from tree and form view

victoralmau:
Add Sign from template action from tree and form view.

Example use case:
- Create a template linked to a model (r`es.partner` for example).
- Go to the partner tree view and select at least one record (or the form view).
- A "_Sign from template_" action will be show automatically.

Only for users with Sign permissions.

ejemplo

@Tecnativa TT41745


victoralmau:
@pilarvargas-tecnativa and @pedrobaeza What do you think about this approach?

sign_oca/security/security.xml
```diff
@@ -38,26 +38,6 @@
         <field name="perm_unlink" eval="0" />
     </record>
     <!-- User Record Rules -->
-    <record id="sign_oca_template_rule_user_read" model="ir.rule">
```
etobella, line 41: Why are you removing the rules?

victoralmau: For many reasons: 
- The current rules are incorrect (the user_id field does not exist, it should be create_uid). 
- Really thinking about it I think it is not necessary to filter templates.

etobella: I see, it seems legit to me!

Thanks!

:+1:

pedrobaeza:
Please rebase for avoiding the conflict. Any screenshot / GIF about what this adds?

victoralmau:
Rebase done + gif added in the initial description with the use case.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):
LGTM thanks!!!

etobella (approved):

PR #21, opened 2024-03-05
[14.0][FIX] sign_oca: Small fixes in configure mode

BernatPForgeFlow:
- Load 'required' field in configure dialog
- Show 'required' fields in red in configure mode
- Load 'role_id' field in configure dialog

PR #17, opened 2024-02-23
[14.0][IMP] sign_oca: preview requests

PauBForgeFlow:
This improvement allows previewing a sign request by clicking the signers counter button. If the request is signed, the preview will show the filled in information, otherwise will show the blank template.

PR #22, opened 2024-03-06
[16.0][IMP] sign_oca: Improvements migrated from 14.0

PauBForgeFlow:
Migrated PRs:
#8 #17 #18 #19 #20 #21 #24 #26 #27 #30


sign_oca/tests/test_sign.py
```diff
@@ -41,6 +41,20 @@ def setUpClass(cls):
...
+            {"name": "Child partner", "parent_id": cls.partner.id}
+        )
+        cls.role_customer = cls.env.ref("sign_oca.sign_role_customer")
+        cls.role_supervisor = cls.env.ref("sign_oca.sign_role_supervisor")
+        cls.role_supervisor.default_partner_id = cls.partner
+        cls.role_child_partner = cls.env["sign.oca.role"].create(
+            {
+                "name": "Child partner",
+                "partner_type": "expression",
+                "expression_partner": "{{object.parent_id.id}}",
```
victoralmau, line 55: Can you add a migration script to convert existing `expression_partner` data?

sign_oca/views/res_partner_views.xml
```diff
@@ -0,0 +1,27 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <div name="button_box" position="inside">
+                <button
+                    class="oe_stat_button"
+                    type="object"
+                    name="action_show_signer_ids"
+                    attrs="{'invisible': [('signer_count', '=', 0)]}"
+                    icon="fa-pencil"
```
victoralmau, line 15:
```suggestion
                    icon="fa-pencil"
                    groups="sign_oca.sign_oca_group_user"
```

sign_oca/views/res_partner_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                >
+                    <field
+                        string="Sign Requests"
+                        name="signer_count"
+                        widget="statinfo"
```
victoralmau, line 21:
```suggestion
                        widget="statinfo"
                        groups="sign_oca.sign_oca_group_user"
```

victoralmau (changes requested):
Thank you for doing this work to v16. 

Some comments (my functional review is still pending).

sign_oca/tests/test_sign.py
PauBForgeFlow: I added the migration script, please check if it works as required.

sign_oca/migrations/16.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def migrate(cr, version):
+    _logger.info("Change the field format.")
+    cr.execute(
+        """
+            UPDATE sign_oca_role
+            SET expression_partner = REPLACE(expression_partner, '${%}', '{{%}}' )
+        """
+    )
```
victoralmau:
```suggestion
from openupgradelib import openupgrade


@openupgrade.migrate()
def migrate(env, version):
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE sign_oca_role
        SET expression_partner = REPLACE(expression_partner, '${%}', '{{%}}' )
        """
    )
```

victoralmau: That's correct, but in the migration scripts use openupgradelib

victoralmau (changes requested):
Some things I have seen do not work in v16:

- The "Sign from template" action that is automatically added in the tree and form view does not work: https://github.com/OCA/sign/pull/20
- Systray management does not work: https://github.com/OCA/sign/pull/8
- The action_signer_sign_url button in the form view of a request should only be shown in the sent state.

Other things I have not been able to check.

victoralmau:
Some comments to keep in mind:

- Do not add commits that are not yet merged in v14.
- Migration script will not be needed (it has been added in https://github.com/OCA/sign/pull/25).
- Systray does not work, it will be necessary to make the same changes as done in https://github.com/OCA/sign/pull/25 (when merged).

victoralmau:
IMO I think considering that there is the PR to migrate to 15.0 https://github.com/OCA/sign/pull/25 with all (I think) these changes contemplated I think the ideal would be to leave it at 15.0 correct, and then bring the changes to 16.0. 
What do you think?

PauBForgeFlow:
> IMO I think considering that there is the PR to migrate to 15.0 #25 with all (I think) these changes contemplated I think the ideal would be to leave it at 15.0 correct, and then bring the changes to 16.0. What do you think?

I am going to review your PR and if everything is OK I will add the migration commit in this PR to carry on the commit history. Is it right for you?

victoralmau:
I had thought of the following.
1- Leave `sign_oca` merged in v15.
2- Continue with this PR as follows:

- Cherry-pick all commits from v15 that are NOT in v16.
- Extra commit of "_adapt_" to v16 (of the previous changes).
- Extra commits of improvements or fixes.

victoralmau (changes requested):
Can you check if you need to add any changes from https://github.com/OCA/sign/pull/25/commits/f4054b68cbd60f04dbe0a0e595565c433b33d128 ?

Please, fix pre-commit errors

victoralmau (approved):

victoralmau:
Ping @pedrobaeza

pedrobaeza:
Why it puts `[WIP]`? Please don't use such title, but use draft state on the PR.

sign_oca/wizards/sign_oca_template_generate_multi.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.env.context.get("model", False)
+
+    model = fields.Char(
+        readonly=True,
+        default=compute_model,
+        # default=lambda self: self.env.context.get('model', False)
+    )
+    template_id = fields.Many2one(
+        comodel_name="sign.oca.template",
+        domain="[('model', '=', model)]",
```
victoralmau, line 16:
```suggestion
        domain="['|', ('model', '=', False),('model', '=', model)]",
```
This is intentional, you can select templates that do not have a defined model ("_global_") or those that are linked to the model.

victoralmau (changes requested):

sign_oca/wizards/sign_oca_template_generate_multi.py
BernatPForgeFlow: Hi @victoralmau,

In our case, we are not using the model functionality as it does not apply to our process. Then it seems a bit too invasive to add a 'Sign From Template' to all models without us being able to use it.

I understand that your idea is to have a generic template that is used to sign multiple models, but maybe we could find an intermediate solution. That is, it would be nice to be able to configure a generic template but only in the case that it is intentional. Also, could you confirm if this is the case or if it's something else so I can finally understand it well?

I honestly don't know what would be the best solution for this. I don't know if it would make more sense for the Model field to be a One2many and here you choose the models you want it to apply to. Or maybe add a checkbox that allows you to mark 'Apply to all models'. What do you think?

victoralmau: Sorry for the misunderstanding, the idea is as follows:

- Only show the action (with js) if there is some template defined for that model.
- In the wizard allow to select templates of the model or without model (the normal will be to use templates of the model).

This behavior is the same as in v15 and I think it will avoid confusion. What do you think?

BernatPForgeFlow: @victoralmau As it works in previous versions, if you have a template with no model the action will appear everywhere. Why would you want to allow selecting templates that does not have a model defined on it?

In context with what you've said, I think that the best solution should be to remove this _model = False_ because adds a non desired behaviour. And then we would only show the action to models that are defined in a template.

victoralmau (approved):

BernatPForgeFlow:
@victoralmau I'm checking the PR but still seem to find something not completely finished. The problem is that when we generate a signature request from the 'Sign from template' action, it doesn't create the signers. It seems to be something that has already been commented here but it's not quite clear to me: https://github.com/OCA/sign/pull/25#issuecomment-2063654757

From my point of view, in the sign_oca code we should generate by default all the signers defined in the template with the active user that is throwing the action. And later, in any case, it could be inherited from the method that generates the signers to define a different logic. So, the 'Sign from Template' feature should work on its own without the need for another module.

Finally, I would also like to ask what exactly is the 'partner_type' field inside the 'sign.oca.role' model? I don't fully understand how to use it. Is it necessary?

victoralmau:
Currently everything is correct although I understand the confusion that exists (perhaps the readme should be improved).

I explain a little bit the different use cases:

**1- Sign_now**
ejemplo--1

A request will be created defining as signer the user that makes it.

**2- Send**
ejemplo-2

A wizard will be created to define the desired signers manually.

**3- Sign from template**
ejemplo-3

A wizard will be created that when selecting a template will generate a request with the "corresponding" signers. The concept of signers does not exist in the templates, it is something that generates from the roles and the type (fixed partner or regular expression). You can check https://github.com/OCA/sign/pull/32 to understand it better.

BernatPForgeFlow:
@victoralmau Ok, I understand how it works and it looks good to me. The only thing that makes me hesitate is that the field is called 'Partner type' as it is not intuitive. How about we change it to 'Partner Selection Policy'? And in addition we add a help explaining a little how it works.

Other than that, what is the use of the 'Empty' option? I don't see any use for it, I would change 'Empty' to 'Active User'. By default, if the role is not configured with another partner selection policy, it would take the active user. Thus, any user can select records from a model and sign them on their behalf.
image

Lastly, when I use the 'Default' option it seems to fail. 
```
Traceback (most recent call last):
  File "/mnt/data/odoo-addons-dir/sign_oca/wizards/sign_oca_template_generate_multi.py", line 35, in generate
    requests = self._generate()
  File "/mnt/data/odoo-addons-dir/sign_oca/wizards/sign_oca_template_generate_multi.py", line 30, in _generate
    return self.env["sign.oca.request"].create(
  File "<decorator-gen-154>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 415, in _model_create_multi
    return create(self, arg)
  File "/mnt/data/odoo-addons-dir/sign_oca/models/sign_oca_request.py", line 298, in create
    records = super().create(vals_list)
psycopg2.ProgrammingError: can't adapt type 'res.partner'
```

victoralmau:
If you want to change the string of the field and add a perfect help, but the empty option should be kept and it is the "_most frequent_".

If we create a `sign.request` with the action "_Sign from template_" from a record (`sale.order` for example) the user that creates it should NOT be auto-added as signer; the ones that correspond with the defined in the template will be set.

The error you indicate I have not been able to reproduce it, can you indicate the steps to reproduce it and/or add a mini-video? If it is something that happens in 15.0 currently it should be solved.

BernatPForgeFlow:
Okay thank you! Will add the changes!

> If we create a `sign.request` with the action "_Sign from template_" from a record (`sale.order` for example) the user that creates it should NOT be auto-added as signer; the ones that correspond with the defined in the template will be set.

About this, why would you add an item in a template that is filled by a role whose partner selection policy is 'Empty'? It's difficult for me to imagine a situation in which you would configure it like that. 
For me, it makes sense that the person who selects the sale order and clicks the action 'Sign from template' is the one who signs the request. It might be useful to validate objects like sale orders, pickings, etc.

A video showing how it is failing for me:
https://github.com/OCA/sign/assets/90243017/6b5d822f-8a63-4130-a9dd-19d70b05a6b6

PauBForgeFlow:
@BernatPForgeFlow @victoralmau The error that you reported has been fixed. Also, the field partner_type has been renamed to partner_selection_policy.

victoralmau (approved):

victoralmau:
> About this, why would you add an item in a template that is filled by a role whose partner selection policy is 'Empty'? It's difficult for me to imagine a situation in which you would configure it like that. For me, it makes sense that the person who selects the sale order and clicks the action 'Sign from template' is the one who signs the request. It might be useful to validate objects like sale orders, pickings, etc.
> 

IMO by default the user who creates it should not be added as a signer. I understand what you comment but there are several things that would not make it very coherent:

- What would be the role for the user (none would be correct).
- If a maintenance user (`maintenance_sign_oca`) for example creates a sign request from a template, it would NOT be correct to set him as signer, you must set the corresponding ones according to the template (the user of the equipment and a fixed user for example).

IMO it could be improved (in a different PR) `sign_oca` so that in the wizard from "_Sign from template_" the signers that will be set with that template are shown, being able to modify them and/or add new ones (for example the user's partner with the desired role).

BernatPForgeFlow (approved):
LGTM! As @victoralmau recommended I will open a separated PR to propose a new way to work with the 'Partner Selection Policy' in 'Sign from template' action.

victoralmau:
Sorry, could you add a migration script to rename field _partner_type_ to _partner_selection_policy_ from `sign.oca.role`?

BernatPForgeFlow:
Ping @pedrobaeza @etobella

sign_oca/migrations/16.0.1.1.1/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 ForgeFlow <http://www.forgeflow.com>
```
pedrobaeza: This should be in pre, not post, or the column will already be created by the ORM and this will give error.

PauBForgeFlow:
@pedrobaeza This is now ready

victoralmau:
Is everything correct now? IMO I think it is important not to delay merging and adding these changes in the 17.0 migration.

pedrobaeza (approved):
/ocabot merge major

PR #28, opened 2024-04-11
[16.0][FIX] sign_oca: sign_oca_pdf checkFilledAll

PauBForgeFlow:
Fix the calculation about if the required items available for the role are filled

Depends on #22


etobella (approved):
/ocabot merge patch

PR #30, opened 2024-04-19
[14.0][FIX+IMP] sign_oca: Minor fixed + improvements

victoralmau:
Changes done:

- [x] Avoid error when deleting an element in the configuration of a request.
- [x] Hide the Sign from template action in `sign.oca*` models
- [x] Improve USAGE section of README

@Tecnativa TT41745


PauBForgeFlow (approved):

sign_oca/readme/USAGE.rst
```diff
@@ -41,3 +41,15 @@ Sign a pending document
 *  There, you can fill all the data you need.
 *  Once you finish, press the sign button on the top
 *  When the last signer signs it, the final file will be generated as a PDF
+
```
pedrobaeza, line 44: No need of 2 lines.

sign_oca/readme/USAGE.rst
```diff
@@ -41,3 +41,15 @@ Sign a pending document
...
+
+Sign from template
+~~~~~~~~~~~~~~~~~~~~~~~
```
pedrobaeza: The underlining should be of the same length.

```suggestion
Sign from template
~~~~~~~~~~~~~~~~~~
```

sign_oca/readme/USAGE.rst
```diff
@@ -41,3 +41,15 @@ Sign a pending document
...
+
+* Go to any list view or form view (except sign.oca models), e.g.: Contacts
+* Select multiple records (3 for example).
+* The "Sign from template" action will be available (it will be shown if there are any sign templates created that are not linked to any model and/or any linked to the corresponding model).
```
pedrobaeza: And please break lines at 88 cols as the same of files.

```suggestion
* The "Sign from template" action will be available if there are any sign templates created that are not linked to any model and/or any linked to the corresponding model.
```

sign_oca/readme/USAGE.rst
```diff
@@ -41,3 +41,15 @@ Sign a pending document
...
+* Select a template.
+* Click on the "Generate" button.
+* 3 requests will be created (each linked to each selected record) BUT no signer will be set.
+* Some extra modules (e.g. maintenance_sign_oca) will automatically set the signers for each request.
```
pedrobaeza, line 55: I think that a good workflow is to:

- Have a method for assigning the signers.
- Default to `user_id` if the field exists in the model, and use it.
- Other modules can override the method to change the signers.

victoralmau: I understand what you want to do (I think that would be another different PR), but signers have `role_id` and `partner_id` fields. Role concept is very important because it can set a default partner or a regular expression, and that is precisely the key to all this (correctly set the template and the fields, you can "_indirectly_" set the signers from request).

pedrobaeza (approved):
/ocabot merge patch

PR #31, opened 2024-04-23
[14.0][FIX] sign_oca: Set signatory data from old format to the new one

BernatPForgeFlow:
We changed signatory data 'role' to 'role_id', so we should change it to old records.

### DIFF
--- a/sign_oca/__manifest__.py
+++ b/sign_oca/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Sign Oca",
     "summary": """
         Allow to sign documents inside Odoo CE""",
-    "version": "14.0.2.4.3",
+    "version": "14.0.2.4.4",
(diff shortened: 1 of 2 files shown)


PauBForgeFlow (approved):
LGTM!

etobella:
I would not change the data, at this point we don't have the chained hash, but that might be a problem if at some point we chain the hashes (it already exists on some versions...).

BernatPForgeFlow:
@etobella Okay, this will only be a problem for people that had been using the Sign app before this commit https://github.com/OCA/sign/commit/45131a27ef7b878d40993773e7e36968d6f6866e since they will have some sign requests with signatory data as 'role' instead of 'role_id'. I will close the PR but leave it as it is just in case someone has the same problem.

PR #34, opened 2024-05-21
[14.0][FIX] sign_oca: Get signer partner info

BernatPForgeFlow:
When a signer signs a request, we will want to fetch the info from his partner instead of fetching from the active user. This allows to load the correct data for signers that do not have a user in the system like customers.

Before:
image

Now:
image


etobella (approved):
:+1:

victoralmau (approved):

PR #37, opened 2024-05-21
[14.0][FIX] sign_oca: Filter correctly next_items

BernatPForgeFlow:

Example:
https://github.com/OCA/sign/assets/90243017/a3504dc9-e728-4fab-8f5e-46180a5c62f2


etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions

PR #41, opened 2024-05-22
[16.0][IMP] sign_oca: Send to signers the final signed document

BernatPForgeFlow:
Once the request is completely signed, all the signers should receive a copy of the final document.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -265,6 +265,40 @@ class SignOcaRequest(models.Model):
                 email_layout_xmlid="mail.mail_notification_light",
             )
 
+    def action_send_signed_request(self):
+        self.ensure_one()
+        if (
+            self.state != "signed"
+            or not self.env.company.sign_oca_send_sign_request_copy
+        ):
+            return
+        for signer in self.signer_ids:
+            attachments = (
+                self.env["ir.attachment"]
+                .sudo()
+                .search(
+                    [
+                        ("res_model", "=", "sign.oca.request"),
+                        ("res_id", "=", self.id),
+                        ("res_field", "=", "data"),
+                    ]
+                )
+            )
+            # The message will not be linked to the record because we do not want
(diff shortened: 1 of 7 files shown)


sign_oca/models/sign_oca_request.py
```diff
@@ -265,6 +265,35 @@ def action_send(self, sign_now=False, message=""):
...
+                .sudo()
+                .search(
+                    [
+                        ("res_model", "=", "sign.oca.request"),
+                        ("res_id", "=", self.id),
+                        ("res_field", "=", "data"),
+                    ]
+                )
+            )
+            self.env["mail.thread"].message_notify(
```
victoralmau, line 289: Can you explain in a code comment why it is not done on the sign request record?

victoralmau (commented):
Code review OK

sign_oca/models/sign_oca_request.py
BernatPForgeFlow: What do you mean it is not done on the sign request record? The code is in the sign request model but we notify all the signers.
What should I exactly clarify?

victoralmau: Sorry, I mean to indicate (so as not to forget it in the future) that the message will not be linked to a record because we don't want it to happen.

BernatPForgeFlow: Okay, done!

victoralmau (approved):

pedrobaeza (commented):
I think this feature should be optional and encode under a configuration option.

BernatPForgeFlow:
> I think this feature should be optional and encode under a configuration option.

@pedrobaeza I added the possibility to configure whether to send signers a copy of the final signed document or not in the settings. It is configured at company level and only visible for sign administrators. WDYT?

pedrobaeza (approved):
Great, this way the feature is there, but not forced for all the users.

/ocabot merge minor

PR #44, opened 2024-05-24
[15.0][FIX] sign_oca: Filter correctly next_items

BernatPForgeFlow:

FWP from 14.0: https://github.com/OCA/sign/pull/37


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #45, opened 2024-05-24
[16.0][FIX] sign_oca: Filter correctly next_items

BernatPForgeFlow:

FWP from 14.0: https://github.com/OCA/sign/pull/37


BernatPForgeFlow:
@pedrobaeza Can we merge this forward port?

pedrobaeza (approved):
/ocabot merge patch

PR #46, opened 2024-05-24
[15.0][IMP] maintenance_sign_oca: Compatibility with hr_maintenance module

victoralmau:
Compatibility with `hr_maintenance` module

If hr_maintenance is installed and the `owner_user_id` field is compute and is not saved, several possible fields in the `write()` must be considered to solve it.

Please @carolinafernandez-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT41745

### DIFF
--- a/maintenance_sign_oca/models/maintenance_equipment.py
+++ b/maintenance_sign_oca/models/maintenance_equipment.py
@@ -75,6 +75,10 @@ class MaintenanceEquipment(models.Model):
         for item in self:
             owners[item.id] = item.owner_user_id
         res = super().write(vals)
-        if vals.get("owner_user_id"):
+        # Fields to be taken into account when trying to create a sign request.
+        # We don't need to take into account only the owner_user_id field because
+        # if you have installed hr_maintenance module is a compute field and the
+        # employee_id field will be taken into account.
+        if any(vals.get(fname) for fname in ["owner_user_id", "employee_id"]):
             self._process_generate_sign_oca_request(owners)
         return res


carolinafernandez-tecnativa (approved):

maintenance_sign_oca/models/maintenance_equipment.py
```diff
@@ -70,11 +70,18 @@ def create(self, vals_list):
             res._process_generate_sign_oca_request({})
         return res
 
+    def _sign_oca_request_get_user_fields(self):
+        """Fields to be taken into account when trying to create a sign request.
+        We don't need to take into account only the owner_user_id field because
+        if you have installed hr_maintenance module is a compute field and the
+        employee_id field will be taken into account."""
+        return [fname for fname in ("owner_user_id", "employee_id") if fname in self]
```
pedrobaeza: Why the if? Can it be directly?

```suggestion
        return ["owner_user_id", "employee_id"]
```

pedrobaeza (commented):
I don't get why having the method `_sign_oca_request_get_user_fields`, the `write` override is not done in the mixin instead of here.

victoralmau:
> I don't get why having the method `_sign_oca_request_get_user_fields`, the `write` override is not done in the mixin instead of here.

It's just a separate method that could be useful if another “_child_” module needed to take another field into account. Would you prefer to remove that method and add it directly in the `write()` method?

pedrobaeza:
I'm saying the contrary: the `write` method shouldn't be needed here, but to do that stuff in the mixin. What is needed is to "instruct" the mixin which fields to look on the write.

victoralmau:
> I'm saying the contrary: the `write` method shouldn't be needed here, but to do that stuff in the mixin. What is needed is to "instruct" the mixin which fields to look on the write.

What mixin? There is no mixin (`sign_oca` or `maintenance_sign_oca`), precisely `maintenance_sign_oca` creates a `sign.request` when a `manintenance.equipment` is created and/or write.

pedrobaeza:
OK, then there's no sense to have a method `_sign_oca_request_get_user_fields` IMO.

victoralmau:
Changes done.

pedrobaeza (approved):
/ocabot merge patch

PR #47, opened 2024-06-21
[16.0][FIX]: translation in owl templates

Demacri:
Load it.po terms in owl frontend templates too

### DIFF
--- a/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.esm.js
+++ b/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.esm.js
@@ -5,6 +5,7 @@ const {App, mount, useRef} = owl;
 import SignOcaPdf from "../sign_oca_pdf/sign_oca_pdf.esm.js";
 
 import env from "web.public_env";
+import core from "web.core";
 import {renderToString} from "@web/core/utils/render";
 import session from "web.session";
 import {templates} from "@web/core/assets";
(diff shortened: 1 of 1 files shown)


pedrobaeza:
If I don't remember bad, what you should do is:

```python
from odoo import models
class IrHttp(models.AbstractModel):
    _inherit = "ir.http"

    @classmethod
    def _get_translation_frontend_modules_name(cls):
        mods = super(IrHttp, cls)._get_translation_frontend_modules_name()
        return mods + ["sign_oca"]
```

pedrobaeza:
You must also follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Demacri:
> If I don't remember bad, what you should do is:
> 
> ```python
> from odoo import models
> class IrHttp(models.AbstractModel):
>     _inherit = "ir.http"
> 
>     @classmethod
>     def _get_translation_frontend_modules_name(cls):
>         mods = super(IrHttp, cls)._get_translation_frontend_modules_name()
>         return mods + ["sign_oca"]
> ```

From what i understand, this avoid to call "session.load_translations" (is it right?), but I'm not able to find a useful documentation about how to handle this scenario in owl components. 

PS: fixed commit guidelines, sorry for the inconvenience.

etobella:
According to Pedro, the change he is asking says:

Return a list of module name where web-translations and
            dynamic resources may be used in frontend views

Actually, some Odoo modules  use it, like payment, portal or im_livechat

PR #55, opened 2024-07-16
[16.0][FIX] sign_oca Fix sign button visibility

### LINKED ISSUE #54: [16.0] `to_sign` being set False incorrectly, sign button hidden from request form
## Module
sign_oca_request

## Describe the bug
Haven't dug into why exactly this is happening, but `to_sign` is False when it shouldn't be, causing the Sign button to be invisible. I just removed the check as a stop gap as the check for `state != sent` does well enough until the `_compute_to_sign` function can be fixed. 

## To Reproduce
**Affected versions**: 
Introduced in [9d9a39f](https://github.com/OCA/sign/commit/e62a782efbd7356fca868a832edc4853e3d1367c)?

1. Create/send sign request
2. Sign button not present in form or tree view

**Expected behavior**
Sign button visible in sign request form and tree views

**Additional context**
Debian 12 with nightly Odoo CE package

ivs-cetmix:
Before this commit
------------------

Internal user cannot sign a document if this user's partner is a child of some other partner.

Example:
User "Pepe Frog" has a partner "Meme Inc, Pepe Frog" set as `user.partner_id`. This is because partner "Meme Inc" is set as a company for the partner "Pepe Frog". So when user Pepe Frog is assigned to a sign request he cannot see the "Sign" button.

After this commit
-----------------

User can see the "Sign" button and sign documents even if related partner has a parent partner set.

This highly likely should close #54

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -362,14 +362,13 @@ class SignOcaRequestSigner(models.Model):
         for item in self.filtered(lambda x: x.request_id.record_ref):
             item.res_id = item.request_id.record_ref.id
 
-    @api.depends("signed_on", "partner_id", "partner_id.commercial_partner_id")
+    @api.depends("signed_on", "partner_id")
     @api.depends_context("uid")
     def _compute_is_allow_signature(self):
         user = self.env.user
         for item in self:
             item.is_allow_signature = bool(
-                not item.signed_on
-                and item.partner_id == user.partner_id.commercial_partner_id
+                not item.signed_on and item.partner_id == user.partner_id
             )
 
     def _compute_access_url(self):


etobella (approved):
Makes sense to me.

Thanks for the fix

sign_oca/models/sign_oca_request.py
```diff
@@ -362,14 +362,13 @@ def _compute_res_id(self):
         for item in self.filtered(lambda x: x.request_id.record_ref):
             item.res_id = item.request_id.record_ref.id
 
-    @api.depends("signed_on", "partner_id", "partner_id.commercial_partner_id")
+    @api.depends("signed_on", "partner_id", "partner_id")
```
pedrobaeza: Remove duplicated depends

```suggestion
    @api.depends("signed_on", "partner_id")
```

ivs-cetmix: Thanks for mentioning that! Fixed)

ivs-cetmix:
Meanwhile another question non related to this PR:
I would like to improve to module documentation too. Can we convert module `/readme` files into markdown for that? It would simply the process a lot)

etobella:
Sure, be my guest

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #56, opened 2024-07-17
[16.0][FIX] sign_oca: Avoid access error with ir.model

victoralmau:
Avoid access error with ir.model

Only "advanced" users have read access to `ir.model`, for this reason, we need to apply `.sudo()`

@Tecnativa


pedrobaeza (approved):
Totally!

/ocabot merge patch

etobella (approved):

PR #58, opened 2024-08-26
[16.0][FIX] sign_oca: The id of the roles must be taken from role_id, not signer_id

etobella:
Otherwise, it does not work properly when signing documents created directly from a request (not coming from a template)


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #68, opened 2024-10-11
[15.0][IMP] sign_oca: Send to signers the final signed document

victoralmau:
Backport from 16.0: https://github.com/OCA/sign/pull/41 + Add General settings menu item

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT41745

PR #69, opened 2024-10-11
[16.0][IMP] sign_oca: Add General settings menu item

victoralmau:
FWP from 15.0: https://github.com/OCA/sign/pull/68

Add General settings menu item

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT41745

### DIFF
--- a/sign_oca/wizards/res_config_settings_views.xml
+++ b/sign_oca/wizards/res_config_settings_views.xml
@@ -33,4 +33,19 @@
             </xpath>
         </field>
     </record>
+    <record id="sign_oca_general_settings_act_window" model="ir.actions.act_window">
+        <field name="name">General Settings</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">res.config.settings</field>
+        <field name="view_mode">form</field>
+        <field name="target">inline</field>
+        <field name="context">{'module' : 'sign_oca'}</field>
+    </record>
+    <menuitem
+        name="General settings"
+        id="sign_oca_general_settings_menu"
+        parent="sign_oca_settings_menu"
+        sequence="1"
+        action="sign_oca_general_settings_act_window"
+    />
 </odoo>


rvalyi (approved):

sign_oca/__manifest__.py
```diff
@@ -14,10 +14,10 @@
         "security/security.xml",
         "views/menu.xml",
         "data/data.xml",
-        "wizards/res_config_settings_views.xml",
         "data/ir_sequence_data.xml",
         "wizards/sign_oca_template_generate.xml",
         "wizards/sign_oca_template_generate_multi.xml",
+        "views/res_config_settings_views.xml",
```
pedrobaeza: Why changing it to `views` and `models`? It's a transient one.

victoralmau: Because even if they are transient `res.config.settings` they have always been set in models and in views as exception (because they are not really a wizard).

pedrobaeza: I would say that's a mistake, but being in the other place, let's not change it.

sign_oca/views/menu.xml
```diff
@@ -13,10 +13,11 @@
         sequence="16"
         groups="sign_oca.sign_oca_group_user"
     />
-    <record model="ir.ui.menu" id="sign_oca_settings_menu">
-        <field name="name">Settings</field>
-        <field name="sequence" eval="90" />
-        <field name="parent_id" ref="sign_oca_root_menu" />
-        <field name="groups_id" eval="[(6,0,[ref('sign_oca.sign_oca_group_admin')])]" />
-    </record>
+    <menuitem
```
pedrobaeza: I don't get why changing this, and more in this commit that is unrelated.

victoralmau: I honestly don't understand these comments, it's something that has already been changed in v15 in the corresponding PR, but no problem, it's already changed.

pedrobaeza (approved):
In v15 it was slipped into the PR, but only for one item, so it was not so critical, but changes should be ordered for the commit history, and we have to avoid cosmetic changes most of the times.

/ocabot merge minor

Issue #71, opened 2024-11-10
[16.0][BUG] when send the same document sign to more signers some tokens are False


anddago78:
When I send a document for signing to multiple signers, an email is sent to each of them, but it seems that only the first recipient sees the button with the correct URL and token in the email. The other signers receive an email with the token set to "false."

## Module
sign_oca

## Describe the bug
No all signers can sign the document...

## To Reproduce
Send a document sign at more signers

Example

![Captura desde 2024-11-10 10-31-55](https://github.com/user-attachments/assets/5780e636-3d90-464a-8ed0-9fc325713388)

![image](https://github.com/user-attachments/assets/d0f18f95-2b68-4b39-ab13-0bae3f5eca88)

![image](https://github.com/user-attachments/assets/0fb95970-ea6b-4ee1-a510-d6dfabe4e295)

This is the link of the second signer....

![image](https://github.com/user-attachments/assets/3697c80a-3ca0-4210-a2ce-d819fb99ec3d)

victoralmau:
Solved with https://github.com/OCA/sign/pull/72

PR #72, opened 2024-11-11
[16.0][FIX] sign_oca: Define the correct url to be set in the email that is sent to the signers so that everyone can sign

victoralmau:
Define the correct url to be set in the email that is sent to the signers so that everyone can sign

Example use case:
- Create a request
- Define 2 different signers
- Click on the Send button
- 2 emails will be sent and each one will have the correct url in the Sign document button

Fixes https://github.com/OCA/sign/issues/71

@Tecnativa


etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

anddago78:
Very thanks @victoralmau for fast resolution! thanks!!

PR #76, opened 2025-01-13
[18.0][IMP] sign_oca: adding events for bus synchronization

kobros-tech:
In this new feature, web client user can observe each sign request update in the status bar without needing to reload the page or even clicking on the pen icon.

Observe that event is sent only to scoped users who should sign on the new created sign request. and not for everybody.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -328,11 +328,24 @@ class SignOcaRequest(models.Model):
             .create(self._set_action_log_vals(action, **kwargs))
         )
 
+    @api.model
+    def notify_changes(self, partner_recs):
+        # send notification to the list of subscribers
+        partner_list_ids = partner_recs.mapped("id")
+        for partner_id in partner_list_ids:
+            self.env["bus.bus"]._sendone(
+                "broadcast",
+                f"sign_oca_request_updates_{partner_id}",
+                {"message": "Sign OCA Requests Model updated"},
+            )
+
     @api.model_create_multi
     def create(self, vals_list):
         records = super().create(vals_list)
         for record in records:
             record._set_action_log("create")
+            partner_recs = record.signer_ids.mapped("partner_id")
+            self.notify_changes(partner_recs)
         return records
 
 
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -331,13 +331,10 @@ class SignOcaRequest(models.Model):
     @api.model
     def notify_changes(self, partner_recs):
         # send notification to the list of subscribers
-        partner_list_ids = partner_recs.mapped("id")
-        for partner_id in partner_list_ids:
-            self.env["bus.bus"]._sendone(
-                "broadcast",
-                f"sign_oca_request_updates_{partner_id}",
-                {"message": "Sign OCA Requests Model updated"},
-            )
+        channel = "sign_oca_request_updates"
+        message = {"message": "Sign OCA Requests Model updated"}
+        for partner_id in partner_recs:
+            partner_id._bus_send(channel, message)
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/sign_oca/__manifest__.py
+++ b/sign_oca/__manifest__.py
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "author": "Dixmit,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sign",
-    "depends": ["web_editor", "portal", "base_sparse_field"],
+    "depends": ["web_editor", "portal", "base_sparse_field", "bus"],
     "data": [
         "security/security.xml",
         "views/menu.xml",
@@ -48,6 +48,7 @@
             "sign_oca/static/src/elements/signature.esm.js",
             "sign_oca/static/src/elements/check.esm.js",
             "sign_oca/static/src/components/sign_oca_pdf/sign_oca_pdf.esm.js",
+            "sign_oca/static/src/js/sign_requests_service.esm.js",
             "sign_oca/static/src/js/sign_oca.esm.js",
             "sign_oca/static/src/js/systray_service.esm.js",
             "sign_oca/static/src/xml/*.xml",
--- a/sign_oca/__manifest__.py
+++ b/sign_oca/__manifest__.py
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "author": "Dixmit,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sign",
-    "depends": ["web_editor", "portal", "base_sparse_field", "bus"],
+    "depends": ["web_editor", "portal", "base_sparse_field"],
     "data": [
         "security/security.xml",
         "views/menu.xml",
--- a/sign_oca/static/src/js/sign_requests_service.esm.js
+++ b/sign_oca/static/src/js/sign_requests_service.esm.js
@@ -1,0 +1,58 @@
+/** @odoo-module **/
+/* global */
+/* Copyright 2025 Kencove - Mohamed Alkobrosli
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+import {Reactive} from "@web/core/utils/reactive";
+import {reactive} from "@odoo/owl";
+import {registry} from "@web/core/registry";
+import {user} from "@web/core/user";
+
+export class WatchSignRequestsService extends Reactive {
+    static modelToLoad = [];
+    static serviceDependencies = ["bus_service", "orm", "notification"];
+
+    constructor() {
+        super();
+        this.ready = this.setup(...arguments).then(() => this);
+    }
+
+    async setup(env, {bus_service, orm, notification}) {
+        this.env = env;
+        this.bus_service = bus_service;
+        this.orm = orm;
+        this.notification = notification;
+        this.sign_requests = reactive({signerCounter: 0, signerGroups: []});
+
+        this.bus_service.subscribe(
+            `sign_oca_request_updates_${user.partnerId}`,
+            async ({message}) => {
+                if (message) {
+                    await this.fetchSystraySigner();
+                }
+            }
+        );
+    }
+    async fetchSystraySigner() {
+        const groups = await this.orm.call("res.users", "sign_oca_request_user_count");
+        let total = 0;
+        for (const group of groups) {
+            total += group.total_records || 0;
+        }
+        this.sign_requests.signerGroups = groups;
+        this.sign_requests.signerCounter = total;
+        return {groups, total};
+    }
+    getTotalSignRequests() {
+        return this.sign_requests.signerCounter;
+    }
+}
+
+export const watchSignRequestsService = {
+    dependencies: WatchSignRequestsService.serviceDependencies,
+    async start(env, services) {
+        return new WatchSignRequestsService(env, services).ready;
+    },
+};
+
+registry.category("services").add("watchSignRequests", watchSignRequestsService);
--- a/sign_oca/static/src/js/sign_requests_service.esm.js
+++ b/sign_oca/static/src/js/sign_requests_service.esm.js
@@ -6,7 +6,6 @@
 import {Reactive} from "@web/core/utils/reactive";
 import {reactive} from "@odoo/owl";
 import {registry} from "@web/core/registry";
-import {user} from "@web/core/user";
 
 export class WatchSignRequestsService extends Reactive {
     static modelToLoad = [];
@@ -24,14 +23,11 @@ export class WatchSignRequestsService extends Reactive {
         this.notification = notification;
         this.sign_requests = reactive({signerCounter: 0, signerGroups: []});
 
-        this.bus_service.subscribe(
-            `sign_oca_request_updates_${user.partnerId}`,
-            async ({message}) => {
-                if (message) {
-                    await this.fetchSystraySigner();
-                }
+        this.bus_service.subscribe("sign_oca_request_updates", async ({message}) => {
+            if (message) {
+                await this.fetchSystraySigner();
             }
-        );
+        });
     }
     async fetchSystraySigner() {
         const groups = await this.orm.call("res.users", "sign_oca_request_user_count");
--- a/sign_oca/static/src/js/systray_service.esm.js
+++ b/sign_oca/static/src/js/systray_service.esm.js
@@ -15,25 +15,13 @@ export class SignerMenuView extends Component {
         this.discussSystray = useDiscussSystray();
         this.orm = useService("orm");
         this.action = useService("action");
+        this.watchSignRequests = useService("watchSignRequests");
         this.state = useState({
-            signerGroups: [],
-            signerCounter: 0,
+            sign_requests: this.watchSignRequests.sign_requests,
         });
         onMounted(async () => {
-            await this.fetchSystraySigner();
+            await this.watchSignRequests.fetchSystraySigner();
         });
-    }
-    async fetchSystraySigner() {
-        const groups = await this.orm.call("res.users", "sign_oca_request_user_count");
-        let total = 0;
-        for (const group of groups) {
-            total += group.total_records || 0;
-        }
-        this.state.signerCounter = total;
-        this.state.signerGroups = groups;
-    }
-    async onBeforeOpen() {
-        await this.fetchSystraySigner();
     }
     availableViews() {
         return [
--- a/sign_oca/static/src/xml/systray.xml
+++ b/sign_oca/static/src/xml/systray.xml
@@ -3,7 +3,6 @@
     <t t-name="sign_oca.SignerMenu">
         <Dropdown
             position="'bottom-end'"
-            beforeOpen.bind="onBeforeOpen"
             manual="false"
             menuClass="discussSystray.menuClass"
         >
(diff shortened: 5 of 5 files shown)


kobros-tech:
@dnplkndll 
@victoralmau
@ovnicraft 
@pedrobaeza
@etobella 

Hi guys, I need you to review this new feature, with which user can observe each sign request update in the status bar without needing to reload the page.

Observe that event is sent only to scoped users and not for everybody.

pedrobaeza:
You should explain the new feature on the commit message and PR description, and if it makes sense in this module or in a separate one.

kobros-tech:
> You should explain the new feature on the commit message and PR description, and if it makes sense in this module or in a separate one.

@pedrobaeza 
no worries this PR is not the new feature that I plan to add in a separate module.

The new one is to show sign requests to customers in the portal and to enable them to sign there.


For this PR feature, it is syncing changes instantly to backend users who are asked to sign and not everybody to save our servers from unnecessarily load.

pedrobaeza:
OK, that explanation is the one it should go in the commit message. Adding 2 new contributors seems too much for this change, isn't it? Then, there are weird changes in .po files that are not coming from your improvement and should be avoided.

kobros-tech:
> OK, that explanation is the one it should go in the commit message. Adding 2 new contributors seems too much for this change, isn't it? Then, there are weird changes in .po files that are not coming from your improvement and should be avoided.

thanks, let me improve it all

kobros-tech:
@pedrobaeza 

I have just added the 3 improvements that you requested from me, I can see the code in the PR is more clean and descriptive.

victoralmau:
Sorry, if I understood the enhancement correctly it should not be necessary to reload the page to update systray, but testing it does not work as expected.
ejemplo

Maybe I am forgetting something?

kobros-tech:
> Sorry, if I understood the enhancement correctly it should not be necessary to reload the page to update systray, but testing it does not work as expected. ejemplo [ ejemplo ](https://private-user-images.githubusercontent.com/4117568/403004147-24a06847-0ffe-4b21-8bbb-b790ebdbd26b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY4NzA0MzgsIm5iZiI6MTczNjg3MDEzOCwicGF0aCI6Ii80MTE3NTY4LzQwMzAwNDE0Ny0yNGEwNjg0Ny0wZmZlLTRiMjEtOGJiYi1iNzkwZWJkYmQyNmIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTRUMTU1NTM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NGMzZTE0YzBiMjFjZTA4YjY5YTliMmFjMzc1YWU3MmI3MGUwODQ3MWFmZTE5MDZkMTI5Mjk1MjUyNTU4OGVjNSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.Od5qbm-cqWysjUsxR2CTBwWpKBdmC1fhgndBNnRMh_A) [ ](https://private-user-images.githubusercontent.com/4117568/403004147-24a06847-0ffe-4b21-8bbb-b790ebdbd26b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY4NzA0MzgsIm5iZiI6MTczNjg3MDEzOCwicGF0aCI6Ii80MTE3NTY4LzQwMzAwNDE0Ny0yNGEwNjg0Ny0wZmZlLTRiMjEtOGJiYi1iNzkwZWJkYmQyNmIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTRUMTU1NTM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NGMzZTE0YzBiMjFjZTA4YjY5YTliMmFjMzc1YWU3MmI3MGUwODQ3MWFmZTE5MDZkMTI5Mjk1MjUyNTU4OGVjNSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.Od5qbm-cqWysjUsxR2CTBwWpKBdmC1fhgndBNnRMh_A)
> 
> Maybe I am forgetting something?

@victoralmau 

all right I will debug this.

could you try to create sign request from button in the template views?

pedrobaeza:
One thing is clear: a total reload of the single-page app shouldn't be done in any case.

kobros-tech:
@victoralmau 
I need you to try this on chrome and on firefox

with me chrome was not listening to bus events, don't know why

kobros-tech:
> One thing is clear: a total reload of the single-page app shouldn't be done in any case.

scenario is not to reload the page at all.

dnplkndll:
> One thing is clear: a total reload of the single-page app shouldn't be done in any case.

are there any other projects working with this concept of pushing the events to the clients over polling for changes? I was hoping to prove it out then move to a more generic approach as you would then be able to push render update events on much more expensive areas to re-render. or areas with a lot of subscribers

kobros-tech:
@pedrobaeza 

Can we figure out whether or no OCA boat is using workers > 0, if they are 0 nobody can use OCA boat for reviewing this PR and will be reviewed locally with --workers=1

pedrobaeza:
@sbidoul should know if runboat can be run with multiple workers activating the bus.

dnplkndll:
> @pedrobaeza
> 
> Can we figure out whether or no OCA boat is using workers > 0, of they are 0 nobody can use OCA boat for review this PR and will be reviewed locally with --workers=1

https://github.com/sbidoul/runboat/blob/main/src/runboat/kubefiles/runboat-start.sh#L39

sbidoul:
runboat runs in multithread mode (the default, --workers=0). Why is that a problem for this PR?

pedrobaeza:
I think the bus doesn't work with multi-threaded mode.

kobros-tech:
> I think the bus doesn't work with multi-threaded mode.

workers should be 1 or more to be effective with this PR or any similar case.

sbidoul:
I don't know what is special in this PR, but I tried chatting between two sessions on the same runboat database and it works, so I would assume the bus is functional on runboat.

I would say that if we have features that requires multi worker, this would need a big warning and/or something that prevent installing the module in multithread mode.

dnplkndll:
since: https://github.com/odoo/odoo/pull/191222 
you would see a warning if not working right?

victoralmau:
Perhaps the way it is used in `mail.activity` can be useful https://github.com/odoo/odoo/blob/81bf1bab9bed9aa0780601d69a1c89942d92b657/addons/mail/models/mail_activity.py#L308

kobros-tech:
> Perhaps the way it is used in `mail.activity` can be useful https://github.com/odoo/odoo/blob/81bf1bab9bed9aa0780601d69a1c89942d92b657/addons/mail/models/mail_activity.py#L308

In fact method used in the link is based on the method I used here, no lower level than what I provided.

I guess chatter is using bus as both methods should use bus module.

sign_oca/models/sign_oca_request.py
```diff
@@ -328,11 +328,24 @@ def _set_action_log(self, action, **kwargs):
...
+    @api.model
+    def notify_changes(self, partner_recs):
+        # send notification to the list of subscribers
+        partner_list_ids = partner_recs.mapped("id")
+        for partner_id in partner_list_ids:
+            self.env["bus.bus"]._sendone(
+                "broadcast",
+                f"sign_oca_request_updates_{partner_id}",
+                {"message": "Sign OCA Requests Model updated"},
+            )
```
nilshamerlinck:
```suggestion
        partner_recs._bus_send(
            "sign.oca.request/insert",
            {}
        )
```

sign_oca/models/sign_oca_request.py
```diff
@@ -328,11 +328,24 @@ def _set_action_log(self, action, **kwargs):
...
+
     @api.model_create_multi
     def create(self, vals_list):
         records = super().create(vals_list)
         for record in records:
             record._set_action_log("create")
+            partner_recs = record.signer_ids.mapped("partner_id")
+            self.notify_changes(partner_recs)
```
nilshamerlinck, line 345:
```suggestion
        partner_recs = records.signer_ids.mapped("partner_id")
        self.notify_changes(partner_recs)
```

sign_oca/models/sign_oca_request.py
nilshamerlinck: 1) better not send to broadcast channel or all users will receive a notification
2) you can use higher-level method `_bus_send()` method on partners (see [here](https://github.com/odoo/odoo/commit/432e375246f0b5ab440d33bb7cfabdb363ef6828))
3) maybe `sign.oca.request` itself could implement `bus.listener.mixin` to cover more events type?

nilshamerlinck (changes requested):

sign_oca/models/sign_oca_request.py
kobros-tech: What I am doing here is a uniqe notification sent to it related partner and not everybody.
Only partners related to the created request are in scope.

If you find something else in review, refer to me, please?

sign_oca/models/sign_oca_request.py
kobros-tech: @ nilshamerlinck 
thanks, but there will be no change in code functionality or syntax.

sign_oca/models/sign_oca_request.py
nilshamerlinck: - by sending your notification to the `broadcast` channel, all clients will receive it
- if you're not convinced, you can inspect messages received by the websocket worker here: `chrome://inspect/#workers`
- of course you are filtering client-side, but server-side would be better (by using the already existing partner channel)

sign_oca/models/sign_oca_request.py
nilshamerlinck: if you create 2 `sign.oca.request` that involve same signers, isn't it better to send only one notification to each signer?

kobros-tech: creating two requests does not make sense.
only one is needed to be fully signed but better if some partner is not there we send him an email again.

we made a PR on that.

kobros-tech: @nilshamerlinck 
#89

victoralmau: Currently it is possible to create multiple requests at the same time, for example, through `maintenance_sign_oca` (although it can be done without this module).

test

victoralmau (changes requested):
Functionally testing this one does not work. Example of another module that does something similar and works correctly: https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/models/tier_validation.py#L784 + https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/static/src/js/services/tier_review_service.esm.js#L14

kobros-tech:
@victoralmau 
The talk here is about sending notification to users of new docs to be signed if they are opening the system, this update can only work with bus port working like --workers=2 when running odoo-bin.

It is not related to creating requests, but it can not be reviewed using OCA boat so better to have a solution of sending notifications without workers app.

@nilshamerlinck gave me suggestion but I need a suggestion to use same port 8069 instead of 8073 of workers, by the way only important notificatios are sent via 8069 but ordinary ones use 8073 port.

victoralmau:
> @victoralmau The talk here is about sending notification to users of new docs to be signed if they are opening the system, this update can only work with bus port working like --workers=2 when running odoo-bin.
> 
> It is not related to creating requests, but it can not be reviewed using OCA boat so better to have a solution of sending notifications without workers app.
> 
Sorry, but I don't understand the reason for having to configure what you indicate, in the previous examples (https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/models/tier_validation.py#L784 + https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/static/src/js/services/tier_review_service.esm.js#L14) it is not necessary to configure anything and it works correctly in oca boat for example.

sign_oca/models/sign_oca_request.py
kobros-tech: ok, it sounds interesting to check

kobros-tech:
> > @victoralmau The talk here is about sending notification to users of new docs to be signed if they are opening the system, this update can only work with bus port working like --workers=2 when running odoo-bin.
> > It is not related to creating requests, but it can not be reviewed using OCA boat so better to have a solution of sending notifications without workers app.
> 
> Sorry, but I don't understand the reason for having to configure what you indicate, in the previous examples (https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/models/tier_validation.py#L784 + https://github.com/OCA/server-ux/blob/59205647d92f9c6c6ee955f1ae23d5f7b8b0409d/base_tier_validation/static/src/js/services/tier_review_service.esm.js#L14) it is not necessary to configure anything and it works correctly in oca boat for example.

this also looks interesting, and I must check

kobros-tech:
@nilshamerlinck 
@victoralmau 

your suggestions were useful to reduce code, it is ok to use one single common channel to send notification as the update will apply on the partners who should sign because large workers need large processors.

I can run the module when workers=1 but not less although I use the same method that you suggest it did not apply with workers=0

victoralmau (changes requested):
IMO it does not work as it should, the icon in the upper right corner with the total number of signatures pending is not updated (it is necessary to reload the page for it to work properly).

ejemplo

kobros-tech:
> IMO it does not work as it should, the icon in the upper right corner with the total number of signatures pending is not updated (it is necessary to reload the page for it to work properly).
> 
> ejemplo [ ejemplo ](https://private-user-images.githubusercontent.com/4117568/457316665-b7401ecb-8620-4b46-99b9-0453d6f399be.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTA2NzcyMjEsIm5iZiI6MTc1MDY3NjkyMSwicGF0aCI6Ii80MTE3NTY4LzQ1NzMxNjY2NS1iNzQwMWVjYi04NjIwLTRiNDYtOTliOS0wNDUzZDZmMzk5YmUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDYyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA2MjNUMTEwODQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YjA5OTBiMDM0NDBhZjdkOGEwYWQ3NGNhOWI5ODc1ZDljNDUxZjkwZDJlYWY0NGU5NmY3NjZlODZiYjY1MTIxMSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.nuN1Fm2wbveQZD9MWz1pOXYU_uOYP9ykLP6O0gsqAaU) [ ](https://private-user-images.githubusercontent.com/4117568/457316665-b7401ecb-8620-4b46-99b9-0453d6f399be.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTA2NzcyMjEsIm5iZiI6MTc1MDY3NjkyMSwicGF0aCI6Ii80MTE3NTY4LzQ1NzMxNjY2NS1iNzQwMWVjYi04NjIwLTRiNDYtOTliOS0wNDUzZDZmMzk5YmUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDYyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA2MjNUMTEwODQxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YjA5OTBiMDM0NDBhZjdkOGEwYWQ3NGNhOWI5ODc1ZDljNDUxZjkwZDJlYWY0NGU5NmY3NjZlODZiYjY1MTIxMSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.nuN1Fm2wbveQZD9MWz1pOXYU_uOYP9ykLP6O0gsqAaU)

Did you test it on run boat? It will not work on run boat unless there is a value for workers > 0.

I applied your suggestion although it needs a value for workers similar to the low level method I used before.

victoralmau:
Yes, it may work in runboat. You can check `base_tier_validation` https://github.com/OCA/server-ux/tree/18.0/base_tier_validation (I initially indicated that module as an example where it is implemented and works correctly). although to check it functionally, you will have to do so with `sale_tier_validation` https://github.com/OCA/sale-workflow/tree/18.0/sale_tier_validation

dnplkndll:
@kobros-tech did this feature with event update over refresh prove valuable? can is be used in other place or in a more general way?

kobros-tech:
> @kobros-tech did this feature with event update over refresh prove valuable? can is be used in other place or in a more general way?

if you plan to use same feature we can think to make a base module, but it needs smart implementation.

not sure but all what I have done I was not sure about before I start.

PR #77, opened 2025-01-16
[18.0][ADD] sign_portal_oca: sign documents in Portal

kobros-tech:
Enable customers, and all portal users of signing documents from Portal.

We still improving it, se we are happy to review and give your opinion.


kobros-tech:
@pedrobaeza

pedrobaeza:
The name in any case should be `sign_portal_oca`, or just `sign_portal` if that name is not used. Does this makes sense to be separated from main module?

sign_oca_portal/views/sign_oca_portal_templates.xml
```diff
@@ -0,0 +1,51 @@
...
+            <t t-if="sign_requests" t-call="portal.portal_table">
+                <thead>
+                    <tr class="active">
+                        <th>
+                            <span class='d-none d-md-inline'>Sign Request #</span>
+                            <span class='d-block d-md-none'>Ref.</span>
+                        </th>
+                        <th class="text-end">To be Signed</th>
+                        <th class="text-end">Partner Name</th>
+                        <th class="text-end">Role</th>
```
victoralmau: I am not sure if it is necessary to show this data, maybe sometimes someone uses names that are not helpful to the customer.

victoralmau (changes requested):
I think it should not be a separate module and could be added directly to `sign_oca`.

The icon is not added in /my to be able to access my/sign_requests.
ejemplo

sign_oca_portal/views/sign_oca_portal_templates.xml
kobros-tech: no problem it is just the introductory, for the icon it should be there.

I don't know why it is not shown, could you install sale module and see.

kobros-tech:
button is shown only if there are documents to be signed

Screenshot from 2025-01-21 18-49-38

pagination is enabled too

Screenshot from 2025-01-21 18-57-07

sign_portal_oca/controllers/portal.py
```diff
@@ -0,0 +1,68 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.http import request, route
+
+from odoo.addons.portal.controllers.portal import pager as portal_pager
+from odoo.addons.sign_oca.controllers.main import PortalSign
+
+
+class PortalSign(PortalSign):
+    def get_sign_requests_domain(self, request):
```
victoralmau, line 11: I think that if in https://github.com/OCA/sign/blob/dfb532e7314027b4ed0911b652b7c8b9caac7dd3/sign_oca/models/res_users.py#L13 you define the domain in a method (`_get_sign_oca_request_user_domain()` for example) you could remove this one and unify it.

sign_portal_oca/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Sign Oca Portal",
+    "summary": """
+        Allow to sign documents inside Odoo Portal CE""",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Kencove, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sign",
+    "depends": ["sign_oca", "portal"],
```
victoralmau, line 12: The `sing_oca` module already has as a `portal` dependency (https://github.com/OCA/sign/blob/dfb532e7314027b4ed0911b652b7c8b9caac7dd3/sign_oca/__manifest__.py#L12) so I think there is no need for an extra module and all this can be added in `sign_oca`.

sign_portal_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,51 @@
...
+            <t t-if="sign_requests" t-call="portal.portal_table">
+                <thead>
+                    <tr class="active">
+                        <th>
+                            <span class='d-none d-md-inline'>Sign Request #</span>
+                            <span class='d-block d-md-none'>Ref.</span>
+                        </th>
+                        <th class="text-end">To be Signed</th>
+                        <!-- <th class="text-end">Partner Name</th> -->
+                        <!-- <th class="text-end">Role</th> -->
```
victoralmau, line 23:
```suggestion
```

sign_portal_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,51 @@
...
+                                t-out="'Yes'"
+                            />
+                            <span t-else="" t-out="'No'" />
+                        </td>
+                        <!-- <td class="text-end">
+                            <span t-field="sign_request.partner_name" />
+                        </td> -->
+                        <!-- <td class="text-end">
+                            <span t-field="sign_request.role_id.name" />
+                        </td> -->
```
victoralmau, line 45:
```suggestion
```

victoralmau (changes requested):
Functional review OK.

sign_portal_oca/__manifest__.py
kobros-tech: what is the value of pagination if I remove portal dependency:
`from odoo.addons.portal.controllers.portal import pager as portal_pager`

victoralmau: Sorry if I have not explained well, the module `sign_oca` already has as a portal dependency, so I commented that in this new module was not necessary and also I commented (as https://github.com/OCA/sign/pull/77#issuecomment-2595942755) that all this should be added directly in the module `sign_oca`.

pedrobaeza: I agree this should be integrated in `sign_oca` module.

sign_portal_oca/views/sign_portal_oca_templates.xml
kobros-tech: done

sign_portal_oca/__manifest__.py
kobros-tech: @victoralmau 
@pedrobaeza 
let me ask Mr. Don @dnplkndll, and if there is a conflict I will make another PR that appends this functionality the the same sign module, are you satisfied now :)

pedrobaeza: We are always satisfied with fresh blood contributing cool things, hehe. It's just to adjust at maximum the fragmentation/modularity for not having split modules when not suitable, or very packed ones when it doesn't fit.

PR #79, opened 2025-01-27
[16.0][ADD] account_avatax_exemption_sign_oca

dnplkndll:
account_avatax_exemption_sign_oca

proposing a new module to streamline the sale tax pdf exemption sign and storage using the sign_oca app and the account_avatax_exemption

we created and used this internal for several years now. Will be adding the account_avatax_exemption_dms next for paper based scan. not in digital signature and then hope to next gave a portal feature allowing the customer to upload or sign request a renewal requirement on a given state sales tax exemption

I wanted to propose the inclusion in oca with very general use case and can then complete test and migrate. 

@JordiBForgeFlow @dreispt @atchuthan


kobros-tech:
@dnplkndll 
They might need us to add a description of the purpose of this module.

I will work on tests

account_avatax_exemption_sign_oca/models/exemption.py
```diff
@@ -0,0 +1,16 @@
...
+class ResPartnerExemption(models.Model):
+    _inherit = "res.partner.exemption"
+
+    sign_oca_request_id = fields.Many2one("sign.oca.request")
+
+    @api.depends("state")
+    def _cancel_sign_request_id(self):
+        for rec in self:
+            if rec.state == "cancel":
+                rec.sign_oca_request_id.cancel()
```
AlexPForgeFlow:
```suggestion
	def write(self, vals):
		if vals.get("state") == "cancel":
			self.sign_oca_request_id.cancel()
		res = super().write(vals)
```
@api.depends only would be triggered to compute fields (don't apply in this field), so I suggest changing the logic to handle cases on write method.

AlexPForgeFlow (changes requested):
Module account_avatax_exemption should be removed from MR as does not belong to this repo.

account_avatax_exemption_sign_oca/models/exemption.py
kobros-tech: thanks, I added this module in a separate PR.

here I added it only because we wanted to test sign exemption and the module for exemption is not merged in 16.0 branch.

we tried to add it from the PR in test requirements as dependency but didn't succeed.

we can resume discussion about exemption module in its open PR for 16.0 branch.

account_avatax_exemption_sign_oca/models/exemption.py
```diff
@@ -1,16 +1,16 @@
...
     sign_oca_request_id = fields.Many2one("sign.oca.request")
 
-    @api.depends("state")
-    def _cancel_sign_request_id(self):
-        for rec in self:
-            if rec.state == "cancel":
-                rec.sign_oca_request_id.cancel()
+    def write(self, vals):
+        if vals.get("state") == "cancel":
+            for record in self:
```
dnplkndll, line 16: @AlexPForgeFlow would you need to call the cascading cancel() in a loop?

AlexPForgeFlow: @dnplkndll You are right! As base cancel method is not handling multi-record https://github.com/OCA/sign/blob/16.0/sign_oca/models/sign_oca_request.py#L233 ,then you have to call cancel method for each record in the loop as your lasts changes.

kobros-tech:
@dnplkndll 
@AlexPForgeFlow 

There were some tests needed update after we modified the functionality under write method.

AlexPForgeFlow (approved):
LGTM! :+1:

kobros-tech (approved):

kobros-tech:
Observe in the exemption view I display the exemption name, and due to:
https://github.com/OCA/account-fiscal-rule/pull/512

I fix an issue with name as there was a treatment with a virtual name, but now now is really exited in the database and editable thanks to the constrain I introduce there.

kobros-tech:
By the way, jut to remind myself and other devs. onchange is concerning view changes ONLY but constrains are not bound to views.

I didn't commit the same of the ` [16.0][ADD] dms_account_avatax_exemption #395 ` here as the module is specific, and the exemption modules themselves are handled in the view by default.

kobros-tech:
@AlexPForgeFlow

what do you think of this commit?
[948d6e8faec5cb9a5ca1db1e653eb408aadd9846](https://github.com/OCA/sign/pull/79/commits/948d6e8faec5cb9a5ca1db1e653eb408aadd9846)

kobros-tech:
I still have one more suggestion to show up the sign_request field in the view if it is there same as what we did in the dms exemption connector so we can go between both views easily and quickily.

AlexPForgeFlow:
@kobros-tech regarding this commit https://github.com/OCA/sign/pull/79/commits/948d6e8faec5cb9a5ca1db1e653eb408aadd9846 .
It's true what you say that onchange methods are for managing changes in the UI, but I do not see any inconvenience on calling them directly (I have seen many implementations calling onchange method directly, also odoo itself). Because at the end, it's a method as all the others, the main difference it's that with tag @api.onchange method will be triggered when a specific field in the UI is changed.
With your proposal, also I see the problem of having duplicated code.

kobros-tech:
> @kobros-tech regarding this commit [948d6e8](https://github.com/OCA/sign/commit/948d6e8faec5cb9a5ca1db1e653eb408aadd9846) . It's true what you say that onchange methods are for managing changes in the UI, but I do not see any inconvenience on calling them directly (I have seen many implementations calling onchange method directly, also odoo itself). Because at the end, it's a method as all the others, the main difference it's that with tag @api.onchange method will be triggered when a specific field in the UI is changed. With your proposal, also I see the problem of having duplicated code.

all right here is the log of the issue in detail:
with on_change line is deleted, yes but also the exemption itself is deleted and the error is raised up, so I changed the UI method with the pure model method.
with my approach exemption is ceated even if the line is removed and the error is no longer raising up!

WRITE called on exemption res.partner.exemption(51,) with vals: {'name': 'Demo Type - 123456 - YourCompany, Mitchell Admin'}
CREATE exemption vals: {'partner_id': 3, 'exemption_type': 19, 'sign_oca_request_id': 24, 'exemption_number': '123456'}
CREATE exemption line vals: {'exemption_id': 51, 'name': 'CERT00023'}
CREATED LINE ID: 24 | exemption_id: 51
==========================================
{'partner_id': 3, 'exemption_type': 19, 'sign_oca_request_id': 24, 'exemption_number': '123456'}
res.partner.exemption(51,)
res.partner.exemption.line(24,)
res.partner.exemption(51,) res.partner.exemption.line(24,) res.partner(3,)
==========================================
WRITE called on exemption res.partner.exemption(51,) with vals: {'business_type': 1}
WRITE called on exemption res.partner.exemption(51,) with vals: {'state_ids': [(6, 0, [])]}
WRITE called on exemption res.partner.exemption(51,) with vals: {'expiry_date': datetime.date(2025, 5, 8)}
WRITE called on exemption res.partner.exemption(51,) with vals: {'exemption_line_ids': [(6, 0, [])]}
WRITE called on exemption line res.partner.exemption.line(24,) with vals: {'exemption_id': False}
2025-04-08 21:59:24,519 107736 ERROR account_avatax_exemption_sign_oca_16_1 odoo.sql_db: bad query: UPDATE "res_partner_exemption_line" SET "exemption_id" = NULL, "partner_id" = NULL, "write_date" = '2025-04-08T21:59:23.699852'::timestamp, "write_uid" = 2 WHERE id IN (24)
ERROR: null value in column "exemption_id" of relation "res_partner_exemption_line" violates not-null constraint
DETAIL:  Failing row contains (24, null, null, null, 1, 2, CERT00023, null, null, t, null, null, 2025-04-08 21:59:23.699852, 2025-04-08 21:59:23.699852).
 
2025-04-08 21:59:24,535 107736 WARNING account_avatax_exemption_sign_oca_16_1 odoo.http: The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model: Avatax Exemption Line (res.partner.exemption.line)
Field: Exemption (exemption_id)

kobros-tech:
@AlexPForgeFlow 
sorry for the line pasting in big format :)

kobros-tech:
@AlexPForgeFlow 
you see now odoo takes both records of exemption and line together and deleted one and the other one that we care about.

That is the side effect of using onchange method, I don't complain about api decorators I use in backend but in this case issue happened.

AlexPForgeFlow:
@kobros-tech When you say the line is deleted, are you referring the exemption lines from a partner? 
Can you give me more details about the issue or steps to reproduce, in order to understand the problem and try to help?

kobros-tech:
This commit is essential for connecting survey, sign, and exemption together:
https://github.com/OCA/sign/pull/79/commits/dbb361f4b511faf5c4bc4c5ef70728f12e589832

kobros-tech:
> @kobros-tech When you say the line is deleted, are you referring the exemption lines from a partner? Can you give me more details about the issue or steps to reproduce, in order to understand the problem and try to help?

you only need to change my methods with on change methods, but you only see this if exemption is created in the backend like when you validate a sign request for exemption.

so using this PR and sign oca you need to interchange current methods and will see an error warning preventing you from validating the sign request.

dnplkndll:
@pedrobaeza what is the process from here on an ADD?

pedrobaeza (changes requested):
I'm struggling with these things:

- The name of the module: it's something more on the sign part, or the Avatax one? That would determine the starting part of the name
- Being any side or the other for the name, Avatax is something very specific of.. USA? Being here in this general repo seems to not be the correct place.
- The README is very short explaining concepts and the purpose of the module. What is Avatax exemption? How is this related to sign?...

dnplkndll:
name:
I was not sure either. the main feature is to store what we need to provide auditors to document the reason a partner is not getting taxed. we have agriculture and resleller. each state has a different version and so you have 100+ templates and then have to store for 3+ years in order to defense an audit ( I did this twice ) 
so there is a requirement to put in emailed and scanned docs into the system to back up the ERP record of the customer taxability type then give in build a given state to the auditor. 
we need a sign and dms system to store all the data and back up the accounting system tax. 
the main work here is sign. nothing really but a link to the exemption crate but needs both. 
country specific:
EE introduced avatax as well. I guess in USA I know here but do not know how needed other places? maybe are simpler to implement than the 1000 counties and state rules on different product classes and changing. it is just too much to implement and then keep up to date. of course that is assuming there are not reliable data feeds on the state or would not. be as hard. 

I am indifferent and can ask on the other project side it appropriate. @dreispt 

readme
@nilshamerlinck we just went over all this could you propose a good readme having just reviewed for first time?

pedrobaeza:
Thanks for the explanation, so, of I understood correctly, this is not something related specifically to Avatax, but tax exemption in general. Or the module contains something tied to the Avatax platform?

dnplkndll:
essentially the purpose if to add the document to support the customer exemption. but that is a good point are the exemptions used by other sales tax systems? I am not sure as to my knowledge we developed this as the first customer sales tax exemption management tool. We should probably be made general purpose to support any sales tax system. However we were not able to use the taxcloud that was the official system on 13.0 as it did not even support customer sales tax exmeptions fort reasons alike agricultural use. I am not certain if that has evened. Do you know of any other systems to review and support in addition to the avatax api?

dnplkndll:
@pedrobaeza oh as to where to put it? just following the maintenance, project convention where is seemed like the sign relates module extensions are in sign. same issue on the PDF supporting document in dms https://github.com/OCA/dms/pull/395 Both depend on an exemption doc to support with a signed doc.

kobros-tech:
@pedrobaeza 

Did you ask to add this module to another project, which one in your opinion?

pedrobaeza:
AFAIK, l10n-usa is hosting the `avatax` modules.

pedrobaeza:
I'm now seeing that the dependent module is in https://github.com/OCA/account-fiscal-rule/tree/16.0/account_avatax_exemption, so there should go this one.

PR #82, opened 2025-02-18
[16.0][FIX] sign_oca: Fixed two issues in signature module

baguenth:
This pull request fixes two issues: 

1) Signature-wizard sometimes does not open

When loading a document to sign using the sign_oca module in Odoo 16, the dialog service is sometimes unavailable during the first page load. This causes an error when attempting to access env.services.dialog. However, after a manual page reload, the issue disappears, indicating a race condition in service initialization.

Root cause:

- The dialog service is initialized asynchronously and may not be available when the SignOcaPdfPortal component is mounted.
- This results in an undefined error when trying to use env.services.dialog.
- The issue is caused by the load order of frontend assets and the delayed initialization of services in Odoo 16.

Solution:

- Implement a polling mechanism using setInterval() to check for the availability of env.services.dialog before mounting SignOcaPdfPortal.
- Once the dialog service is available, stop the polling (clearInterval()) and proceed with the component initialization.

2) Signature-field is sometimes overlapped by other elements

In some cases (based on the pdf content) the signature field is overlapped by the pdf content. This can be fixed by giving the signature field the highest possible z-level.

### DIFF
--- a/sign_oca/static/src/scss/sign.scss
+++ b/sign_oca/static/src/scss/sign.scss
@@ -16,6 +16,8 @@
         border: transparent;
         background-color: transparent;
     }
+
+    z-index: 2147483647; // highest possible css zindex to ensure that the field is never overlapped
 }
 .o_sign_oca_field_config {
     position: absolute;
--- a/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.esm.js
+++ b/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.esm.js
@@ -8,6 +8,7 @@ import env from "web.public_env";
 import {renderToString} from "@web/core/utils/render";
 import session from "web.session";
 import {templates} from "@web/core/assets";
+
 export class SignOcaPdfPortal extends SignOcaPdf {
     setup() {
         super.setup(...arguments);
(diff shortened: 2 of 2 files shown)


baguenth:
Kindly asking for a review :) 

@etobella
@pedrobaeza

sign_oca/static/src/scss/sign.scss
```diff
@@ -16,6 +16,8 @@
         border: transparent;
         background-color: transparent;
     }
+
+    z-index: 2147483647; // highest possible css zindex to ensure that the field is never overlapped
```
pedrobaeza, line 20: Isn't anything that can be on top of this?

baguenth: Technically this is the highest possible z-index. If an element in the pdf comes with the same z-index we may still have an issue. But I guess the chance for this situation is fairly small.

pedrobaeza: Sorry, I didn't mean that. What I'm saying is if this should be indeed the highest possible one, or is there any possible element that can be most on top (due to business needs or so)?

domschmidt: Hi @pedrobaeza,

I'm responding on behalf of @baguenth as the original contributor of this PR (https://github.com/AmetrasIntelligence/sign/pull/1).

The click listener for the signature field only works if the signature field is reachable.

A simple PDF created in Microsoft Word can already cause PDF content to have a higher z-index than the signature field. As a result, the click listener of the signature field is not triggered, and consequently, the dialog does not open.

However, a dialog—such as the actual signature dialog—is opened outside the PDF iframe. Even if the underlying signature field has the highest z-index, the dialog itself appears on an even higher level. Therefore, the workflow should be back in order.

Best regards

pedrobaeza: Thanks for the explanations.

pedrobaeza (approved):

baguenth:
@etobella can you may take a look? :)

etobella (approved):
Looks well explained to me.

PR #83, opened 2025-02-18
[16.0][FIX] sign_oca: Prevent duplicate dialogs per item

baguenth:
Prevent Multiple Dialog Openings for the Same Item

Description

This PR addresses an issue in the sign_oca module where multiple dialogs could be opened for the same item, leading to redundant UI elements and potential user confusion. The issue was further exacerbated by a request delay on:

🔗 http://.../web/sign/get_fonts/

To resolve this, an item-specific guard (item.dialogOpened) has been implemented to prevent duplicate dialog openings for the same item while still allowing different items to open their respective dialogs concurrently.

Changes Made

✅ Added an item-level flag (item.dialogOpened) to prevent duplicate dialogs.
✅ Ensured dialogs reset properly when closed.
✅ Prevented multiple openings due to the request delay on /web/sign/get_fonts/.
✅ Ensured no impact on concurrent dialogs for different items.

Steps to Reproduce the Issue (Before Fix)

Open a signature field in the sign_oca module.
Rapidly trigger the dialog opening action (e.g., clicking multiple times).
Observe that multiple dialogs are opened for the same item.

Expected Behavior (After Fix)

✔️ Only one dialog can be opened per item at a time.
✔️ Different items can still open their own dialogs.
✔️ The fix handles request delays gracefully and prevents unintended duplicate openings.

### DIFF
--- a/sign_oca/static/src/elements/signature.esm.js
+++ b/sign_oca/static/src/elements/signature.esm.js
@@ -22,29 +22,36 @@ const signatureSignOca = {
             core.qweb.render("sign_oca.sign_iframe_field_signature", {item: item})
         )[0];
         if (item.role_id === parent.info.role_id) {
+            const requestOpenDialog = () => {
+                if (!item.dialogOpened) {
+                    item.dialogOpened = true;
+                    var signatureOptions = {
+                        fontColor: "DarkBlue",
+                        defaultName: parent.info.partner.name,
+                    };
+                    parent.env.services.dialog.add(
+                        SignatureDialog,
+                        {
+                            ...signatureOptions,
+                            uploadSignature: (data) =>
+                                this.uploadSignature(parent, item, signatureItem, data),
+                        },
+                        {
+                            onClose: () => {
+                                item.dialogOpened = false;
+                            },
+                        }
+                    );
+                }
+            };
+
             signatureItem[0].addEventListener("focus_signature", () => {
-                var signatureOptions = {
-                    fontColor: "DarkBlue",
-                    defaultName: parent.info.partner.name,
-                };
-                parent.env.services.dialog.add(SignatureDialog, {
-                    ...signatureOptions,
-                    uploadSignature: (data) =>
-                        this.uploadSignature(parent, item, signatureItem, data),
-                });
+                requestOpenDialog();
             });
             input.addEventListener("click", (ev) => {
                 ev.preventDefault();
                 ev.stopPropagation();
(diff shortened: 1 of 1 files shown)


baguenth:
Kindly asking for a review :) 

@etobella
@pedrobaeza

sign_oca/static/src/elements/signature.esm.js
```diff
@@ -22,29 +22,36 @@ const signatureSignOca = {
...
+                    parent.env.services.dialog.add(
+                        SignatureDialog,
+                        {
+                            ...signatureOptions,
+                            uploadSignature: (data) =>
+                                this.uploadSignature(parent, item, signatureItem, data),
+                        },
+                        {
+                            onClose: () => {
+                                item.dialogOpened = false;
```
pedrobaeza, line 41: Isn't this variable need to be initialized in the lifecycle for avoiding a situation where the dialog is never opened having a previous true value?

domschmidt: Hi @pedrobaeza,

Thank you for your question. I'm responding on behalf of @baguenth, as I am the original contributor of this PR (https://github.com/AmetrasIntelligence/sign/pull/2).

The variable is initialized lazily, which is technically fine as long as no other part of the code relies on the same variable for a different purpose.

Additionally, since we are not using typed JavaScript (such as TypeScript), there are no safeguards to prevent new variables from being defined dynamically. This means that, in theory, someone could introduce unintended behavior, but within the current implementation, this should not cause any issues.

Of course, as a maintainer, you have the authority to enforce changes if you believe an explicit initialization would improve code clarity or prevent potential pitfalls. Let me know if you'd like any adjustments!

pedrobaeza (approved):

sign_oca/static/src/elements/signature.esm.js
pedrobaeza: I think it's OK this way, thanks.

baguenth:
@etobella can you may take a look? :)

etobella (approved):

baguenth:
@etobella Thank you very much for your approval. Is there something we need to wait for before we can merge?

baguenth:
@etobella Is there something hindering us from merging?

baguenth:
@etobella @pedrobaeza Here is another merge request, since the pull request is ready to merge for a few weeks now :)

domschmidt (approved):

PR #85, opened 2025-03-03
[16.0][FIX] sign_oca: list requests from recent sent to oldest signed

kobros-tech:
If we have a thousand of requests, it is quite hard to check them as they are listed in ascending way.

In real world we need to inspect not signed and top recent requests.
In the same way no need to see signed requests or canceled requests at the top of the list, but at the end instead.

Screenshot from 2025-03-04 01-40-43
Screenshot from 2025-03-04 01-21-52

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -25,6 +25,7 @@ class SignOcaRequest(models.Model):
     _name = "sign.oca.request"
     _inherit = ["mail.thread", "mail.activity.mixin"]
     _description = "Sign Request"
+    _order = "sequence, create_date desc, id desc"
 
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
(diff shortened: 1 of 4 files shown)


sign_oca/models/sign_oca_request.py
```diff
@@ -25,6 +25,7 @@ class SignOcaRequest(models.Model):
     _name = "sign.oca.request"
     _inherit = ["mail.thread", "mail.activity.mixin"]
     _description = "Sign Request"
+    _order = "sequence, write_date desc, create_date desc, id desc"
```
pedrobaeza: `write_date` should never be a criteria to order by, as this can be modified by a lot of reasons.

kobros-tech: I like your idea, and I will see how to improve

kobros-tech:
@pedrobaeza 
Shall we approve and merge?

kobros-tech:
@dnplkndll
@wlin-kencove

dnplkndll:
could we allow the create date to be optionally shown? 

is the sent type available here? I know we have email and SMS options to send. not sure if there is a type we can see when sending again.

pedrobaeza:
@etobella should say if he considers the approach of having a computed sequence as a good one.

PR #78, opened 2025-01-24
[18.0][IMP] sign_oca: enable signing from portal

kobros-tech:
adding a new feature by enabling portal users from signing their documents from portal in addition to signing it from emails too.
#86

PR #86, opened 2025-03-05
[16.0][IMP] sign_oca: enable signing from portal

kobros-tech:
adding a new feature by enabling portal users from signing their documents from portal in addition to signing it from emails too.
same as #78 but for version 16.0


kobros-tech:
dnplkndll

flotho (approved):
thanks
tested on runboat,
LGTM

sign_oca/controllers/main.py
```diff
@@ -104,3 +104,62 @@ def get_sign_oca_sign_access(
...
+            ("partner_id", "child_of", [request.env.user.partner_id.id]),
+            ("signed_on", "=", False),
+        ]
+        return domain
+
+    def _prepare_sign_portal_rendering_values(self, page=1, sign_page=False, **kwargs):
+        SignRequests = request.env["sign.oca.request.signer"].sudo()
+        domain = self.get_sign_requests_domain(request)
+
+        self._items_per_page = 10
```
xaviedoanhduy:
```suggestion
        self._items_per_page = 10
```
This variable should be declared outside the function. Or it is possible to just declare the local variable type in the current function.

xaviedoanhduy (approved):
LGTM about functionality and code review

sign_oca/tests/test_sign_portal.py
```diff
@@ -71,3 +105,103 @@ def test_portal(self):
...
             data[key] = val
+
+    def test_get_sign_requests_domain(self):
+        self.domain = self.portal_sign.get_sign_requests_domain(request=self)
+        expected_domain = [
+            ("request_id.state", "=", "sent"),
+            ("partner_id", "child_of", [self.env.user.partner_id.id]),
+            ("signed_on", "=", False),
+        ]
+        self.assertEqual(sorted(self.domain), sorted(expected_domain))
```
victoralmau: Personally I do not like this test, I think it would be more appropriate to make a portal test with a tour that checks that a user x when accessing the portal page can see the request previously created in setUp().

Example: https://github.com/OCA/rma/blob/b02f7ddceedca9ffc8b7e8ad44fe3cae315d82aa/rma_sale/tests/test_rma_sale_portal.py#L54

sign_oca/tests/test_sign_portal.py
```diff
@@ -71,3 +105,103 @@ def test_portal(self):
...
+            (
+                "odoo.addons.sign_oca.controllers.main.PortalSign"
+                "._prepare_home_portal_values"
+            ),
+            capture._prepare_home_portal_values,
+        ):
+            response = self.portal_sign._prepare_home_portal_values(
+                counters=["sign_oca_count"]
+            )
+            self.assertEqual(response.home_values, expected_home_values)
```
victoralmau: I think it is not necessary to test all possible methods or use Mock, it would not be enough to do a portal test with a tour.

sign_oca/controllers/main.py
victoralmau: I believe that this should not be modified and the value should be left as it is already defined,

sign_oca/views/portal_templates.xml
```diff
@@ -0,0 +1,47 @@
...
+
+    <template
+        id="portal_my_home"
+        inherit_id="portal.portal_my_home"
+        priority="100"
+        name="Portal layout : sign_requests button"
+        customize_show="True"
+    >
+        <xpath expr="//div[hasclass('o_portal_docs')]" position="inside">
+            <div class="o_portal_category row g-2 mt-3" id="portal_sign_category">
```
victoralmau: What is the reason for wanting to add this extra div that doesn't leave everything together?
ejemplo

sign_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,43 @@
...
+                            <a t-att-href="sign_request.get_portal_url()">
+                                <t t-out="sign_request.request_id.name" />
+                            </a>
+                        </td>
+                        <td class="text-end">
+                            <span
+                                t-if="sign_request.request_id.to_sign"
+                                t-out="'Yes'"
+                            />
+                            <span t-else="" t-out="'No'" />
```
victoralmau: I think this is a bit redudant, that is, if it is shown in the listing it is because you can sign it, otherwise it will not be shown. Perhaps the `_display_name` of the linked object could be shown (if it is).

victoralmau (changes requested):

sergiocorato (approved):
Thanks

sign_oca/views/portal_templates.xml
kobros-tech: I will find a way to make it smooth

sign_oca/views/sign_portal_oca_templates.xml
kobros-tech: I guess we will continue to show aleady signed docs in the portal, we will need it
but I am not sure yet to resume here or delay it for another PR

victoralmau: Ok, if you resume it here, go ahead, but otherwise, better remove this unnecessary and inconsistent column with the “Your Documents to be Signed” text currently displayed.

victoralmau (changes requested):
My other comments are pending: https://github.com/OCA/sign/pull/86#discussion_r1993783421 + https://github.com/OCA/sign/pull/86#discussion_r1993772625 + https://github.com/OCA/sign/pull/86#discussion_r1993776223 + https://github.com/OCA/sign/pull/86#discussion_r1993783421

kobros-tech:
> My other comments are pending: [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993783421) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993772625) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993776223) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993783421)
@victoralmau 

There are mock tests and tour tests already there, you can review!
If you don't like number of rows I added as 10, what is the default should be there?
The new breadcrup should be now within others smoothly, so please review agin these issues properly.

Any other ones that could be related to new work will be handled according to what we decide whether remving or keeping to resume work.

I expect to stop at this point as the commit is getting large and review will never stop, so I tend to remove the lement for upcomming contribution.

Thanks ;)


could you guide me if I add my tour how can user open a page and return back to open another one like mine.
I think I tried but I got timeout.

victoralmau:
> > My other comments are pending: [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993783421) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993772625) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993776223) + [#86 (comment)](https://github.com/OCA/sign/pull/86#discussion_r1993783421)
> > @victoralmau
> 
> There are mock tests and tour tests already there, you can review! If you don't like number of rows I added as 10, what is the default should be there? The new breadcrup should be now within others smoothly, so please review agin these issues properly.
> 
> Any other ones that could be related to new work will be handled according to what we decide whether remving or keeping to resume work.
> 
> I expect to stop at this point as the commit is getting large and review will never stop, so I tend to remove the lement for upcomming contribution.
> 
> Thanks ;)

Please comment on each thing in the corresponding conversation, if you want to. You can also skip my review comments, of course.

sign_oca/views/portal_templates.xml
kobros-tech: fixed

kobros-tech:
@victoralmau 
I asked you if you don't like this default value to show:
self._items_per_page = 10

which is the other default one? as in the method I set the value by the way 
I choose how many lines to appear, if you have an idea I didn't catch up explain more and better to have an example with code.

https://github.com/odoo/odoo/blob/18.0/addons/portal/controllers/portal.py#L133
https://github.com/odoo/odoo/blob/18.0/addons/hr_timesheet/controllers/portal.py#L77

For tests we have tour test and methods tests, including mocking if you don't like mock then it is a solution better than none.
tour is good but functionality for each part is good to be tested too.

We do mocking in avatax not just here.

I see the PR is very ready now

sign_oca/controllers/main.py
victoralmau: This definition is not necessary (`sale` for example does not define it), when it is not defined the value of the `portal` module (https://github.com/odoo/odoo/blob/4cb362fcc3c2ba2489ef1de7c20c9a4fd8921ab8/addons/portal/controllers/portal.py#L133) will be used, something that in most cases is enough. It will only be re-defined when you want to change the “_default_” value (`hr_timesheet` for example), but I think `sign_oca` does not need to change it.

sign_oca/tests/test_sign_portal.py
victoralmau: I think this test is not necessary, because it is testing “_not necessary_” things, all this is already being tested with the portal test (_test_sign_requests_) which is defined later.

kobros-tech:
@etobella 
@victoralmau 
@pedrobaeza 

You have made entire change on the module by deleting many files and a dir, it is ok to remove vunrable code but now I have conflict and I have to rebuld my work from zero as long as WE want it to be merged into OCA.

So, can we have a serious look into the open PRs and prepare for merging, I am ready for it and waiting.

Then, I can fix the scroll PR with the conflict.

pedrobaeza:
It's true that #105 removes code, but it reduces a lot of duplicated code, don't you think? I recommend you also to subscribe to the repository to be aware of the current pull requests, for commenting about before merging, so the rest can be aware.

kobros-tech:
> It's true that #105 removes code, but it reduces a lot of duplicated code, don't you think? I recommend you also to subscribe to the repository to be aware of the current pull requests, for commenting about before merging, so the rest can be aware.

all right, I would like to know how to subscribe to a repository in OCA?

pedrobaeza:
Here you have it:

imagen

kobros-tech:
@pedrobaeza 

I really need to have this module merged as long as there is no more issues to solve.

between time to time I find conflicts with the base branch and I have to fix the conflict, we still have a migration to 18.0 of this feature and the PR is closed although it is a concerning feature, right?

https://github.com/OCA/sign/pull/77

kobros-tech:
@pedrobaeza

kobros-tech:
@pedrobaeza

victoralmau:
Again, the same comment as https://github.com/OCA/sign/pull/86#issuecomment-2750339259 

My comments (https://github.com/OCA/sign/pull/86#discussion_r1993783421 + https://github.com/OCA/sign/pull/86#discussion_r1993772625 + https://github.com/OCA/sign/pull/86#discussion_r1993776223 + https://github.com/OCA/sign/pull/86#discussion_r1993783421) are still pending (even though they have been marked as resolved).

kobros-tech:
@pedrobaeza 
@victoralmau 

now the tests are simpler and the default number is set, it should be 80 by the way.

Can we have a progress in review and hopefully to get merged?

kobros-tech:
@pedrobaeza

kobros-tech:
@pedrobaeza 
@victoralmau 

I pushed and squashed the commit, it is ready after applying your 3 comments, resolved now?

sign_oca/controllers/main.py
```diff
@@ -104,3 +104,60 @@ def get_sign_oca_sign_access(
         return signer_sudo.action_sign(
             items, access_token=access_token, latitude=latitude, longitude=longitude
         )
+
+    def get_sign_requests_domain(self, request):
+        domain = [
+            ("request_id.state", "=", "sent"),
+            ("partner_id", "child_of", [request.env.user.partner_id.id]),
+            ("signed_on", "=", False),
```
victoralmau:
```suggestion
            ("request_id.state", "in", ("sent", "signed")),
            ("partner_id", "child_of", [request.env.user.partner_id.id]),
```
IMO this should be the appropriate domain for displaying all pending and signed requests.

The texts on the portal should probably be more consistent to take this into account.

victoralmau (changes requested):

kobros-tech:
@victoralmau 

done

kobros-tech:
Because we updated the state value and merged this update, I had to apply changes for state values here.

sign_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,42 @@
...
+                        <td>
+                            <a t-att-href="sign_request.get_portal_url()">
+                                <t t-out="sign_request.request_id.name" />
+                            </a>
+                        </td>
+                        <td class="text-end">
+                            <span
+                                t-if="sign_request.request_id.to_sign"
+                                t-out="'Yes'"
+                            />
```
victoralmau, line 65:
```suggestion
                            <span
                                t-if="sign_request.request_id.to_sign"
                                t-out="'Yes'"
                            />
                            <span
                                t-else=""
                                t-out="'No'"
                            />
```

victoralmau: If something has already been signed, a text will also be displayed.

victoralmau (changes requested):
Functional review OK

sign_oca/views/sign_portal_oca_templates.xml
kobros-tech: we would allow them to download the document also.

kobros-tech:
@victoralmau 
@pedrobaeza 

any suggestions, or go to merge?

pedrobaeza:
I need @etobella confirmation.

victoralmau (approved):

etobella (changes requested):
It is quite confusing that the text is "Your Documents to Be Signed". I have signed it but still says the same....

Image

Image

On the list I would add:
1- Signature Date (If there is)
2- Clicking on signed documents should not open the signed document, should download the signed document....
3- I would recomment to see the log report on documents
4- I would prefer to see a column state than a "To Be Signed" column.

etobella:
BTW, I really like the idea and approach of showing this data in the portal, but just a few comments before we are ready (I will be fast on reviewing)

kobros-tech:
image

image

kobros-tech:
@etobella 
@pedrobaeza 

Now portal users can download signed documents, any suggestions?

once OK, I will squash into 1 commit.

etobella:
Ok, a few things:

1. Signature Date as the aggregates is  not correct. It should be Signed Documents or Signature REquests.
2. I would recomment to add filters on that view similar to sale. At least the sort and filter for state would be awesome.
    
image

3. When downloading the reports, it does not add the proper name of the filename. It is just called data (without extension)

kobros-tech:
@etobella 

I am not used to most of OCA modules, can you pass to me the links to modules are doing filters in portal and etc.
If you have an example of what you asked me for I may understand quickly.

kobros-tech:
@etobella 

I guess the 3 requests of yours are implemented now, I hpe?

Screenshot from 2025-09-04 00-31-23

kobros-tech:
I don't want someone to ask me to sort based on state value, it will cause me to rebult the entire code of this PR :)

kobros-tech:
image

kobros-tech:
image

kobros-tech:
When we agree on the second commit changes, I will add test cases as I realized tests are very important especially when we migrate databases from version to version.

etobella:
This is a sort, It would be nice to see a filter. You have examples in the module project, where you can see the tasks in the portal.

kobros-tech:
Screenshot from 2025-09-18 20-09-08

@etobella 
here it is

sign_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,70 @@
...
+                                            class="mx-2"
+                                            title="Not fully signed!"
+                                        >⏳</span>
+                                    </a>
+                                </t>
+                            </t>
+                        </td>
+                        <td class="text-start">
+                            <!-- t-field is essential to show display name not technical name -->
+                            <span t-out="sign_request.signed_on" />
```
etobella: <span
                                    t-field="sign_request.signed_on"
                                    t-options='{"widget": "datetime"}'
                                />

kobros-tech: Thanks it really helps :)

kobros-tech:
@etobella 
@pedrobaeza 

Are we ready now?

pedrobaeza:
@etobella you tell us

etobella:
When I download the file from here: 

image


The file is downloaded without an extension.

image

sign_oca/controllers/main.py
```diff
@@ -104,3 +108,125 @@ def get_sign_oca_sign_access(
...
+            .sudo()
+            .search(
+                [
+                    ("res_model", "=", "sign.oca.request"),
+                    ("res_id", "=", sign_request.id),
+                    ("res_field", "=", "data"),
+                ],
+                limit=1,
+            )
+        )
```
victoralmau, line 217: IMO, I think an `attachment_id` store=False field should be created with this domain in `sign.oca.request.signer `to simplify the code https://github.com/OCA/sign/blob/16.0/sign_oca/models/sign_oca_request.py#L289

Furthermore, I believe that the `portal_download_signed()` method is unnecessary because it could be used for the download /web/content/{attachment_id} url (something that the `web` allows: https://github.com/odoo/odoo/blob/215ad0e1c7d74f919c646e40ed23c0bb12cee034/addons/web/controllers/binary.py#L65).

kobros-tech: @victoralmau 

I did not understand what do you mean in the first sentence, can you explain more the difference and the requested change, please?

kobros-tech:
@victoralmau 

return request.redirect(
            f"/web/content/{attachment.id}?download=true&filename={utf8_filename}"
        )
        
 using this line is very nice but the module was not compatable with it, if I use it I need access_token value for the attachment or for the sign.oca.request.signer or sign.oca.request but sign.oca.request has no field access_token but has attachment
 
 and sign.oca.request.signer has access_token but does not have attachment

xaviedoanhduy:
> using this line is very nice but the module was not compatable with it, if I use it I need access_token value for the attachment or for the sign.oca.request.signer or sign.oca.request but sign.oca.request has no field access_token but has attachment

hi @kobros-tech, you can get `access_token` from `generate_access_token()` function from attachment created above 
https://github.com/odoo/odoo/blob/e8d4873ecf27d10399b67b62a03cce6c47a2f69b/addons/mass_mailing/models/mailing.py#L1404

kobros-tech:
@xaviedoanhduy 
@victoralmau 
@pedrobaeza 
@etobella 

now the suggestions are added and no issue with access, thanks xaviedoanhduy

kobros-tech:
hello? 
@etobella

sign_oca/views/sign_portal_oca_templates.xml
```diff
@@ -0,0 +1,73 @@
...
+                </thead>
+                <t t-foreach="sign_requests" t-as="sign_request">
+                    <tr>
+                        <td>
+                            <t t-if="sign_request.request_id.state == '2_signed'">
+                                <a
+                                    t-att-href="'/my/sign/%s/download' % sign_request.request_id.id"
+                                >
+                                    <t t-out="sign_request.request_id.name" />
+                                    <span class="mx-2" title="Download">💾</span>
```
etobella: 1. This is not needed anymore.
2. If you needed to use a save button, you should have used https://fontawesome.com/v4/icon/floppy-o

kobros-tech: @etobella 

I remember I tried before but these favicons were not available in odoo 16, I used other good ones too.

kobros-tech:
I am waiting for celebrating the merge!

etobella (approved):

Issue #87, opened 2025-03-05
sign_oca: send button we just need to resend versus create new


kobros-tech:
send button we just need to resend versus create new

PR #89, opened 2025-03-05
[16.0][IMP] sign_oca: add resend feature to partners who did not sign

kobros-tech:
Sometimes we can resend the sign request for some partners and find that a new one is created although some people have already signed the previous one.

In that case we will have to resign agin which is extra needed

This PR is allow to resend the email or sms of the same sign request to be signed, also there is an option to allow creating request for once.

#87 

unsubscribe feature is added to avoid sending emails to partners who are not interested.

Only partners who didn't sign and still followers to the specific sign request are receiving the email notifications.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -255,23 +255,9 @@ class SignOcaRequest(models.Model):
         self._set_action_log("validate")
         self.state = "0_sent"
         for signer in self.signer_ids:
-            signer._portal_ensure_token()
             if sign_now and signer.partner_id == self.env.user.partner_id:
                 continue
-            render_result = self.env["ir.qweb"]._render(
-                "sign_oca.sign_oca_template_mail",
-                {"record": signer, "body": message, "link": signer.access_url},
-                engine="ir.qweb",
-                minimal_qcontext=True,
-            )
-            self.env["mail.thread"].message_notify(
-                body=render_result,
-                partner_ids=signer.partner_id.ids,
-                subject=_("New document to sign"),
-                subtype_id=self.env.ref("mail.mt_comment").id,
-                mail_auto_delete=False,
-                email_layout_xmlid="mail.mail_notification_light",
-            )
+            signer.action_send(sign_now, message)
 
     def action_send_signed_request(self):
         self.ensure_one()
@@ -445,6 +431,27 @@ class SignOcaRequestSigner(models.Model):
             "type": "ir.actions.act_url",
             "url": self.access_url,
         }
+
+    def action_send(self, sign_now=False, message=""):
+        self.ensure_one()
+        if self.signed_on or (sign_now and self.partner_id == self.env.user.partner_id):
+            return
+        self._portal_ensure_token()
+        render_result = self.env["ir.qweb"]._render(
+            "sign_oca.sign_oca_template_mail",
+            {"record": self, "body": message, "link": self.access_url},
+            engine="ir.qweb",
+            minimal_qcontext=True,
+        )
+        # send the message to partners who didn't sign only
+        self.env["mail.thread"].message_notify(
+            body=render_result,
+            partner_ids=self.partner_id.ids,
+            subject=_("New document to sign"),
+            subtype_id=self.env.ref("mail.mt_comment").id,
+            mail_auto_delete=False,
+            email_layout_xmlid="mail.mail_notification_light",
+        )
 
     def action_sign(self, items, access_token=False, latitude=False, longitude=False):
         self.ensure_one()
--- a/sign_oca/views/sign_oca_request.xml
+++ b/sign_oca/views/sign_oca_request.xml
@@ -111,6 +111,14 @@
                         <tree editable="bottom">
                             <field name="role_id" />
                             <field name="partner_id" />
+                            <button
+                                string="Send"
+                                type="object"
+                                attrs="{'invisible': [('signed_on', '!=', False)]}"
+                                name="action_send"
+                                icon="fa-paper-plane"
+                                confirm="Are you sure ?"
+                            />
                             <field name="signed_on" />
                         </tree>
                     </field>
@@ -306,6 +314,15 @@
                                             widget="kanban_activity"
                                         />
                                     </div>
+                                    <button
+                                        string="Send"
+                                        type="object"
+                                        states="1_draft,0_sent"
+                                        name="action_send"
+                                        icon="fa-paper-plane"
+                                        confirm="Are you sure ?"
+                                        class="btn btn-primary btn-sm"
+                                    />
                                 </div>
                                 <div class="oe_kanban_bottom_right">
                                     <field
@@ -331,6 +348,14 @@
                 <field name="request_id" />
                 <field name="signed_on" />
                 <field name="altered_hash" invisible="1" />
+                <button
+                    string="Send"
+                    type="object"
+                    attrs="{'invisible': [('signed_on', '!=', False)]}"
+                    name="action_send"
+                    icon="fa-paper-plane"
+                    confirm="Are you sure ?"
+                />
             </tree>
         </field>
     </record>
@@ -345,6 +370,14 @@
                         type="object"
                         name="sign"
                         attrs="{'invisible': [('is_allow_signature', '=', False)]}"
+                    />
+                    <button
+                        string="Send"
+                        type="object"
+                        attrs="{'invisible': [('signed_on', '!=', False)]}"
+                        name="action_send"
+                        icon="fa-paper-plane"
+                        confirm="Are you sure ?"
                     />
                 </header>
                 <sheet>


kobros-tech:
### Notification of success if email is resent successfully

Screenshot from 2025-03-06 12-43-00
Screenshot from 2025-03-06 12-47-30
Screenshot from 2025-03-06 12-40-50

kobros-tech:
### Resend button is shown in the kanban view:

Screenshot from 2025-03-06 13-07-00

kobros-tech:
@flotho

xaviedoanhduy (changes requested):
hi @kobros-tech, why don't you change or reuse the `action_send` ([ref](https://github.com/OCA/sign/blob/f3989f2a848233f77ab9f5453335a4cc1bfef04c/sign_oca/models/sign_oca_request.py#L250)) function to make the `sign.oca.request` support for `sent` state as well instead of creating a duplicate function with the same functionality. same as the **Send By Email** function of Sale Order.
image

kobros-tech:
> hi @kobros-tech, why don't you change or reuse the `action_send` ([ref](https://github.com/OCA/sign/blob/f3989f2a848233f77ab9f5453335a4cc1bfef04c/sign_oca/models/sign_oca_request.py#L250)) function to make the `sign.oca.request` support for `sent` state as well instead of creating a duplicate function with the same functionality. same as the **Send By Email** function of Sale Order. image

if it could be applied would be a good idea💡

kobros-tech:
@victoralmau

victoralmau (commented):
I don't understand the need to resend the email, in any case, maybe it would be easier to add a “Back to draft” button (while in sent status) to achieve the same goal. What do you think?

kobros-tech:
> I don't understand the need to resend the email, in any case, maybe it would be easier to add a “Back to draft” button (while in sent status) to achieve the same goal. What do you think?

The point is not creaing many requests as if there are 3 partners and they should sign 
Two already signed but one still not, we need to notify him again if we send the default is to create a new doc request if he signs the other two should also sign agin on the new created request.

So we better resend the same signed request with 2 partners to him to sign, so the point is not the state at all but the workflow.

kobros-tech:
by the way I didn't see back to draft button here is it recently merged or what?

kobros-tech:
Only partners who have not signed yet will receive the email to sign.

kobros-tech:
I am planning to correct the module to prevent creation if the signers are the same, return to draft button is not shown to me in this module?

xaviedoanhduy (approved):
> I am planning to correct the module to prevent creation if the signers are the same, return to draft button is not shown to me in this module?

I agree with your point of view. We should make the function of sending mail similar to that of sending email like Sale Order, Purchase Order, Invoice, ... (we can send mail multiple times in draft, sent or even confirm status)
Your change now is ok with me, so thanks for your work.

dnplkndll:
> I don't understand the need to resend the email, in any case, maybe it would be easier to add a “Back to draft” button (while in sent status) to achieve the same goal. What do you think?

The use case would be to remind the signer that there is a pending item needing attention. not sure if there is any better way to accomplish that. I guess if they are not longer interested in the service/status achieved from the contract, then the email should have a cancel option to prevent additional reminders. 
@kobros-tech can the signer/email recipient indicate they would like to not sign/cancel the request?

kobros-tech:
> > I don't understand the need to resend the email, in any case, maybe it would be easier to add a “Back to draft” button (while in sent status) to achieve the same goal. What do you think?
> 
> The use case would be to remind the signer that there is a pending item needing attention. not sure if there is any better way to accomplish that. I guess if they are not longer interested in the service/status achieved from the contract, then the email should have a cancel option to prevent additional reminders. @kobros-tech can the signer/email recipient indicate they would like to not sign/cancel the request?

technically yes, they can unsubscribe to a channel

but let me try if I can.

kobros-tech:
@dnplkndll 
unsubscribe can be for sign_oca all or for resending notification only, so which one I shoud take?

kobros-tech:
### Unsubscribe Feature:

Screenshot from 2025-05-12 20-27-08

Screenshot from 2025-05-12 20-27-23

kobros-tech:
> I don't understand the need to resend the email, in any case, maybe it would be easier to add a “Back to draft” button (while in sent status) to achieve the same goal. What do you think?

there is an idea to achieve the same goal, shall I try and if it works we prefer the best option?
I can make PR for the other approach and after discussion and review we select the best one?
@dnplkndll

sign_oca/controllers/main.py
```diff
@@ -104,3 +104,32 @@ def get_sign_oca_sign_access(
...
+            return {"state": "wrong_access"}
+        return {"state": "ok"}
+
+    @http.route("/mail/unsubscribe", type="http", auth="public", website=True)
+    def unsubscribe(self, signer_id=None, res_id=None, access_token=None, **kwargs):
+        """unsubscription controller"""
+        vals = self.unsubscription(
+            signer_id=signer_id, res_id=res_id, access_token=access_token, **kwargs
+        )
+        return request.render("sign_oca.page_unsubscribed", vals)
```
victoralmau: I don't understand the need for this. I mean, why would anyone want to unsubscribe from something they have to sign up for? Do we want to “allow” them to do so? I don't think so. I could apply the same “logic” to an email for a sales order, and no unsubscribe option is added, but that's just my opinion.

sign_oca/models/sign_oca_template.py
```diff
@@ -24,6 +24,7 @@ class SignOcaTemplate(models.Model):
     model = fields.Char(compute="_compute_model", compute_sudo=True, store=True)
     active = fields.Boolean(default=True)
     request_ids = fields.One2many("sign.oca.request", inverse_name="template_id")
+    create_single_request = fields.Boolean(default=True)
```
victoralmau: I don't understand the need for this field; I think it just complicates the (clear) logic that currently exists.

sign_oca/views/sign_oca_request.xml
```diff
@@ -20,7 +20,7 @@
                     <button
                         string="Send"
                         type="object"
-                        states="1_draft"
+                        states="1_draft,0_sent"
```
victoralmau: Continuing with https://github.com/OCA/sign/pull/124#discussion_r2329296336, this is NOT what I suggested. The “send” button in the header should remain as it is and should not be changed. A button (an icon that is representative without text) should be added to the signers, and that action should send the email to that recipient, nothing else.

victoralmau: I think the “_simple_” thing to do is to separate this part https://github.com/OCA/sign/blob/e1440502718c02f3363c88f679dcc5ade01baaf7/sign_oca/models/sign_oca_request.py#L244 into a specific method in sign.request. signer and call that method to “resend.”

victoralmau (changes requested):

kobros-tech:
@etobella 
@pedrobaeza

sign_oca/models/sign_oca_request.py
```diff
@@ -446,6 +432,27 @@ def sign(self):
...
+        # send the message to partners who didn't sign only
+        if not signer.signed_on:
+            self.env["mail.thread"].message_notify(
+                body=render_result,
+                partner_ids=signer.partner_id.ids,
+                subject=_("New document to sign"),
+                subtype_id=self.env.ref("mail.mt_comment").id,
+                mail_auto_delete=False,
+                email_layout_xmlid="mail.mail_notification_light",
+            )
```
victoralmau:
```suggestion
        self.ensure_one()
        if self.signed_on or (sign_now and self.partner_id == self.env.user.partner_id):
            return
        self._portal_ensure_token()
        render_result = self.env["ir.qweb"]._render(
            "sign_oca.sign_oca_template_mail",
            {"record": self, "body": message, "link": self.access_url},
            engine="ir.qweb",
            minimal_qcontext=True,
        )
        # send the message to partners who didn't sign only
        self.env["mail.thread"].message_notify(
            body=render_result,
            partner_ids=self.partner_id.ids,
            subject=_("New document to sign"),
            subtype_id=self.env.ref("mail.mt_comment").id,
            mail_auto_delete=False,
            email_layout_xmlid="mail.mail_notification_light",
        )
```

sign_oca/models/sign_oca_request.py
```diff
@@ -250,28 +250,14 @@ def open_template(self):
 
     def action_send(self, sign_now=False, message=""):
         self.ensure_one()
-        if self.state != "1_draft":
+        if self.state not in ["1_draft", "0_sent"]:
```
victoralmau:
```suggestion
        if self.state != "1_draft":
```

sign_oca/views/sign_oca_request.xml
victoralmau:
```suggestion
                        states="1_draft"
```

sign_oca/views/sign_oca_request.xml
```diff
@@ -111,6 +111,13 @@
                         <tree editable="bottom">
                             <field name="role_id" />
                             <field name="partner_id" />
+                            <button
+                                string="Send"
+                                type="object"
+                                attrs="{'invisible': [('signed_on', '!=', False)]}"
+                                name="action_send"
+                                icon="fa-paper-plane"
```
victoralmau, line 119:
```suggestion
                                icon="fa-paper-plane"
                                confirm="Are you sure ?"
```

sign_oca/views/sign_oca_request.xml
```diff
@@ -353,6 +375,13 @@
                         <field name="role_id" readonly="1" />
                         <field name="partner_id" readonly="1" />
                         <field name="request_id" readonly="1" />
+                        <button
+                            string="Send"
+                            type="object"
+                            attrs="{'invisible': [('signed_on', '!=', False)]}"
+                            name="action_send"
+                            icon="fa-paper-plane"
+                        />
```
victoralmau: This should be added to the header of this view.

sign_oca/models/sign_oca_request.py
victoralmau: I think a note should also be added to `sign.oca.request` to clarify that this process has been carried out, otherwise the “Send” button may be pressed several times because it appears that nothing is happening.

victoralmau (changes requested):

dnplkndll:
@kobros-tech when someone asks for the request but it goes to spam or not received for other reason ( bad email ?) 
could we get link to send through the DM / chat you are talking to the customer on?

sign_oca/views/sign_oca_request.xml
```diff
@@ -23,6 +23,7 @@
                         states="1_draft"
                         name="action_send"
                         icon="fa-paper-plane"
+                        confirm="Are you sure ?"
```
victoralmau:
```suggestion
```
I believe that this specific confirmation is not necessary.

victoralmau (changes requested):
My comment https://github.com/OCA/sign/pull/89#discussion_r2366846548 is still pending.

kobros-tech:
> @kobros-tech when someone asks for the request but it goes to spam or not received for other reason ( bad email ?) could we get link to send through the DM / chat you are talking to the customer on?

I created this PR so that it is very easy to DM the link in the chat or someone can sign on behalf:

https://github.com/OCA/sign/pull/128

kobros-tech:
@victoralmau 

hello?

victoralmau:
My comment https://github.com/OCA/sign/pull/89/files#r2374878936 is still pending.

kobros-tech:
@etobella

kobros-tech:
@victoralmau 

the comment is resolved, is there any comment?

dnplkndll:
@kobros-tech could a plan be setup? for instance to resend in an hour, then each day for a week then monthly unless the signer cancels?  just for instance. that seems kind of annoying too.

kobros-tech:
we can make it in a private custom code, or I  am working on notification automation and I would make it based on queue project.

to approach your point as long as any sign request is set to sent a queued notification will be sent untill the request becomes canceled or signed.

It is still needing finishing but can be reviewed to suggest any features.

https://github.com/OCA/server-tools/pull/3425

victoralmau (approved):

kobros-tech:
@pedrobaeza 
@etobella 

hello, I almost forgot this PR

kobros-tech:
> we can make it in a private custom code, or I am working on notification automation and I would make it based on queue project.
> 
> to approach your point as long as any sign request is set to sent a queued notification will be sent untill the request becomes canceled or signed.
> 
> It is still needing finishing but can be reviewed to suggest any features.
> 
> [OCA/server-tools#3425](https://github.com/OCA/server-tools/pull/3425)

@pedrobaeza thanks very much, if you can review this new module or ping someone to help me getting it through, please?

I have 3 new modules in server-tools project and I do not know who can help me through.

PR #15, opened 2023-12-03
[IMP] sign_oca: Ensure inalterability

etobella:
This change should allow us to make an inalterable signature item, making it much easier for us.

With `sign_oca` and `sign_biometric_oca` we should be able to have and advanced signature following eIDAS, making it legal

PR #93, opened 2025-03-20
[FIX] oca_sign: migrate SECUR_SIGN

dnplkndll:


victoralmau (changes requested):
I'm sorry but I don't understand why this is necessary. The sequence was added at https://github.com/OCA/sign/pull/15 and even if it is a noupdate=1 data, it will be created when the module is updated.

kobros-tech:
> I'm sorry but I don't understand why this is necessary. The sequence was added at #15 and even if it is a noupdate=1 data, it will be created when the module is updated.

I very large databases, we need to make sure the record is created!
If it is not created by update then the migration script should fix this flow.

kobros-tech:
### Migrate ir.sequence record:
2025-03-23 14:42:23,938 12273 INFO sign_oca_16_1 odoo.modules.migration: module sign_oca: Running migration [>16.0.2.1.5] pre-migration 
2025-03-23 14:42:24,049 12273 INFO sign_oca_16_1 OpenUpgrade: sign_oca: pre-migration script called with version 16.0.2.0.2 
2025-03-23 14:42:24,071 12273 INFO sign_oca_16_1 pre-migration: ir.sequence(17,) record is created with ID: 17, with ref: sign_oca.sign_inalterability_sequence 

### Fix this warinin in the log:

2025-03-23 13:52:22,118 9851 WARNING sign_oca_16_1 odoo.addons.base.models.ir_model: Two fields (model, model_id) of sign.oca.template() have the same label: Model. [Modules: sign_oca and sign_oca]

kobros-tech:
@dnplkndll 
@victoralmau 
@AlexPForgeFlow

AlexPForgeFlow:
@dnplkndll @kobros-tech @victoralmau @etobella 
Check this PR https://github.com/OCA/sign/pull/94 where I explain a bit the root cause of issue and my proposal

PR #99, opened 2025-04-02
[ADD] survey_sign_oca: survey and sign_oca connector

kobros-tech:
Because of the nice odoo survey interaction and organized questions and well designed format, there is a need to show the signers a survey and fill, then send them a filled sign request waiting for their approval so that we save time for them and fetch information in an interesting way not a boring one.

The commit is under development and we need to apply the idea in a professional way, and business bearing one!

**description, configuration is TODO**

based on PR 
#91 

#102 

https://github.com/OCA/survey/#157


dnplkndll:
@kobros-tech I think you want to make a new module for this connector. 

I think there needs to be a complete process review too. if the fields are to map from the sign to the survey should that be able to be created from the sign template? 

maybe we could use the US streamlines as a demo item since it covers the most US states and would be able to then create the exemption needed for the tex service. 

https://www.streamlinedsalestax.org/docs/default-source/forms/exemption-certificateb926a7ab4a0d43e1ad4fe8eb19e79cbb.pdf?sfvrsn=857843d_5

kobros-tech:
Screenshot from 2025-04-03 01-55-48

kobros-tech:
https://github.com/user-attachments/assets/cf9f8335-63ad-431e-9dab-f02bed05c4da

kobros-tech:
> @kobros-tech I think you want to make a new module for this connector.
> 
> I think there needs to be a complete process review too. if the fields are to map from the sign to the survey should that be able to be created from the sign template?
> 
> maybe we could use the US streamlines as a demo item since it covers the most US states and would be able to then create the exemption needed for the tex service.
> 
> https://www.streamlinedsalestax.org/docs/default-source/forms/exemption-certificateb926a7ab4a0d43e1ad4fe8eb19e79cbb.pdf?sfvrsn=857843d_5

now I improved the field item to be just one "Survey" and according to each one of them I compare the placehloder to the key of the answer, video is showing that.

I didn't want to give my suggestion of creating a new module untill you give me the suggetion on it.

I will try to apply other ones too.

etobella (changes requested):
Create a glue module. do not add the dependancy here

victoralmau (changes requested):
In addition to https://github.com/OCA/sign/pull/99#pullrequestreview-2738584660, I believe the module could be similar to [maintenance_sign_oca](https://github.com/OCA/sign/tree/16.0/maintenance_sign_oca). Besides, it seems that what is commented in https://github.com/OCA/sign/pull/91#issuecomment-2754775538 will also be necessary, right?

kobros-tech:
> In addition to [#99 (review)](https://github.com/OCA/sign/pull/99#pullrequestreview-2738584660), I believe the module could be similar to [maintenance_sign_oca](https://github.com/OCA/sign/tree/16.0/maintenance_sign_oca). Besides, it seems that what is commented in [#91 (comment)](https://github.com/OCA/sign/pull/91#issuecomment-2754775538) will also be necessary, right?

yes, it's better to have it in a separate module, I wish I could as I have to patch some js methods.

etobella:
We will check the hooks and add them if they have sense. Do not worry about that.

However, I think we shouldn't add unnecessary dependancies on the module.

kobros-tech:
Now I add a standalone module to connect between sing_oca and survey.

kobros-tech:
@dnplkndll

kobros-tech:
If there is a signatur field in the request it will throw error, and it is handle in this PR #102 it is inside sign_oca module.

kobros-tech:
Screenshot from 2025-04-06 08-50-34

Screenshot from 2025-04-06 08-50-49

kobros-tech:
https://youtu.be/KaGNmiYn7ZM

survey_sign_oca/models/sign_oca_request.py
```diff
@@ -0,0 +1,83 @@
...
+    def format_answer(answer):
+        if SurveyUtils.is_yes_no_answer(answer):
+            answer = SurveyUtils.answer_is_yes(answer)
+        return answer
+
+
+class SignOcaRequest(models.Model):
+    _inherit = "sign.oca.request"
+
+    # This field is required for the inverse of maintenance.equipment.
```
victoralmau:
```suggestion
    # This field is required for the inverse of survey.user.input.
```

survey_sign_oca/models/survey_invite.py
```diff
@@ -0,0 +1,73 @@
...
+        self.ensure_one()
+        result = self.env["ir.actions.act_window"]._for_xml_id(
+            "sign_oca.sign_oca_request_act_window"
+        )
+        result["domain"] = [("id", "in", self.sign_request_ids.ids)]
+        ctx = dict(self.env.context)
+        ctx.update(
+            {
+                "default_maintenance_equipment_id": self.id,
+                "search_default_maintenance_equipment_id": self.id,
```
victoralmau:
```suggestion
                "default_maintenance_equipment_id": self.id,
                "search_default_maintenance_equipment_id": self.id,
```
These fields are not the correct ones, you will have to change them for the corresponding ones.

survey_sign_oca/models/survey_survey.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 Kencove - Mohamed Alkobrosli
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class Survey(models.Model):
+    _inherit = "survey.survey"
+
+    start_button_name = fields.Char("Survey Start Button Name", default="Start")
```
victoralmau: Why is this field necessary for sign_oca? What use case are you trying to cover? Is this really necessary in this module?

survey_sign_oca/models/sign_oca_request.py
```diff
@@ -0,0 +1,83 @@
...
+                elif survey.get(placeholder) and item["field_type"] == "check":
+                    item["value"] = SurveyUtils.format_answer(survey.get(placeholder))
+                elif survey.get(placeholder) and item["field_type"] == "signature":
+                    item["value"] = survey.get(placeholder)
+        return vals
+
+    def get_info(self, access_token=False):
+        vals = super().get_info(access_token)
+        vals = self.fill_survey_related_items(vals)
+        return vals
```
victoralmau: Can you explain what this is intended to do (adding comments in the methods explaining what they do would help). 
Are you trying to fill in survey data through the signature?

survey_sign_oca/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Kencove - Mohamed Alkobrosli
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Survey Sign Oca",
+    "version": "16.0.1.0.0",
+    "category": "Surveys",
+    "website": "https://github.com/OCA/sign",
+    "author": "Kencove, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["sign_oca", "survey", "survey_question_type_binary"],
```
victoralmau:
```suggestion
    "depends": ["sign_oca", "survey", "survey_question_type_binary"],
```
IMP the `survey_question_type_binary` dependency should not be necessary.

victoralmau (changes requested):

survey_sign_oca/models/survey_invite.py
kobros-tech: It's a big mistake, I will fix

survey_sign_oca/models/survey_survey.py
kobros-tech: The module is extending the concept meaning of survey module, now it can take a rool for many other functions like filling the exemption data but not in the document directly, so the case is not exactly answering a survey.

So survey module can do jobs like: survey, questionaire, filling docs and maybe evenmore.

`Start Survey` button name is very specific, and it can be start a survey, stary a questionaire start a foo.

survey_sign_oca/models/sign_oca_request.py
kobros-tech: yes, I will

I am just sending the same data from the server to the sign_request frontend so signers will see their answers in the corresponding field places.

survey_sign_oca/__manifest__.py
kobros-tech: it is important if the survey takers will give answers of type binary or multibinary or signature.

It is very important for them to sign, and I can not ignore this field as I need it, and it is available in an OCA module.

kobros-tech: but I will try to use get method if the field is present or not, I will try no problem.

kobros-tech:
@victoralmau 
for the https://github.com/OCA/sign/pull/91#issuecomment-2754775538 commit if you are not going to merge, I really need the field_type to be shown in the form view and allow users to deal with. because if someone delete the data I introduce they can not create it again!.

kobros-tech:
@etobella 
@victoralmau 

I removed the dependency from manifest, but still having the external ones. The one related to survey is not critical here and optional, but the one related to sign_oca is important if getting merged otherwise I show the field I want here instead.

Some handy comments are added to, changes are in a separate commit for simplicity for now.

once you agree we transition to test cases.

survey_sign_oca/models/survey_survey.py
victoralmau: I understand the functionality, the issue is that I think this functionality (specific start and end button texts in surveys) should not be in this specific module, if needed, you should create a module that extends surveys just for that functionality (`survey_fill_form_button_name` or something like that). This is just my opinion.

kobros-tech:
@wlin-kencove

https://github.com/OCA/survey/pull/180

kobros-tech:
@victoralmau 

Is there anymore comments?

survey_sign_oca/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2025 Kencove - Mohamed Alkobrosli
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="name">res.config.settings.view.form.inherit</field>
+        <field name="model">res.config.settings</field>
+        <field name="priority" eval="93" />
+        <field name="inherit_id" ref="base.res_config_settings_view_form" />
```
victoralmau, line 9: The proper thing to do would be to use the Survey > Settings option which does not exist, using a new block I think is not the most appropriate, IMO there are 2 possible options:

A- Add those fields to Sign > General settings, inheriting from `sign_oca.res_config_settings_view_form`.
B- Create a `base_survey_config` module similar to https://github.com/OCA/maintenance/tree/16.0/base_maintenance_config to add a Survey > Settings menu and have this module depend on the other one.

I think the easiest option in this case is option A.

victoralmau (commented):
Code review OK

kobros-tech:
@etobella 
@victoralmau 

how about this module?

PR #100, opened 2025-04-03
Update sign_oca_pdf_portal.xml

anasassi119:


etobella (changes requested):
Why are you enforcing the height?

Also, the commit message is wrongly formed

anasassi119:
For logos that are large, the logo can take up all the screen and there is nothing that's preventing it from doing that. Enforced a height for the logo so that it won't stay in front of the PDF file.

etobella:
But the same might happen in the invoices, isn't it?

anasassi119:
In invoices the height is limited by the container dimensions. Here, it is not limited by anything other than the logo size. This PR isn't just because I read the code, I tried it. Try it yourself, a larger logo would ruin the interface and will cause the signing area to be very small. Enforcing a height for the logo itself will solve the issue.

anasassi119:
Well, it did solve the issue. The 
image
image

anasassi119:
Another issue is that it doesn't support Arabic encoding, but that's a problem for another day...

etobella:
Ok, but you shouldn't use height, as it enforces to enlarge if the logo is smaller. you should use: max-height

And change the commit name to follow the guidelines please.

About the arabic display, it is something to review, but, as you said, probably on a different PR.

PR #101, opened 2025-04-03
[16.0][IMP] maintenance_sign_oca: correct the form of role expression

kobros-tech:
If I follow the same syntax of the configure section I get error, the syntax should be correct so that users don't get messed out.

I need this PR to be merged in other versions of the same issue.


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #102, opened 2025-04-03
[FIX] sign_oca: correct syntax and handle js logic

kobros-tech:
Role correct syntax is between {{ }} not ${}, this will cause error. parent.items[nextItem.id] if is not defined the function dispatchEvent will throw error in the UI, so it is better to handle that error.

### DIFF
--- a/sign_oca/static/src/elements/check.esm.js
+++ b/sign_oca/static/src/elements/check.esm.js
@@ -46,9 +46,12 @@ const checkSignOca = {
             ).sort((a, b) => a.tabindex - b.tabindex);
             if (next_items.length > 0) {
                 ev.currentTarget.blur();
-                parent.items[next_items[0].id].dispatchEvent(
-                    new Event("focus_signature")
-                );
+                const nextItem = next_items[0];
+                if (nextItem && parent.items && parent.items[nextItem.id]) {
+                    parent.items[nextItem.id].dispatchEvent(
+                        new Event("focus_signature")
+                    );
+                }
             }
         });
         return input;
--- a/sign_oca/static/src/elements/signature.esm.js
+++ b/sign_oca/static/src/elements/signature.esm.js
@@ -14,7 +14,10 @@ const signatureSignOca = {
             (i) => i.tabindex > item.tabindex
         ).sort((a, b) => a.tabindex - b.tabindex);
         if (next_items.length > 0) {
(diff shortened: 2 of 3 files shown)


AlvaroRM11 (approved):
LGTM

kobros-tech:
@victoralmau

please help me with this...

pedrobaeza (changes requested):
Please don't mix things in the same PR. I'm OK with the README change. If you publish it isolatedly, we can merge. But the other one is not clear how do you get to a situation where the element is not defined, and even if valid, you shouldn't log by console the warning. Just don't act.

kobros-tech:
> Please don't mix things in the same PR. I'm OK with the README change. If you publish it isolatedly, we can merge. But the other one is not clear how do you get to a situation where the element is not defined, and even if valid, you shouldn't log by console the warning. Just don't act.

just open my #99 and add a signature field when you fill it the error message will pop up, it is an error in UI not in the log.

and in my PR there is no JS there!

pedrobaeza:
As said, please split both changes in 2 PRs for immediate merge of the README one, and the other shouldn't log to console.

kobros-tech:
> As said, please split both changes in 2 PRs for immediate merge of the README one, and the other shouldn't log to console.

All right I will split them

kobros-tech:
> As said, please split both changes in 2 PRs for immediate merge of the README one, and the other shouldn't log to console.

@pedrobaeza 

#104  the readme one

etobella (approved):
For me it has sense

pedrobaeza:
Enric, also the console.warn part?

etobella:
Sorry, I didn't read your comment about the log :thinking:

Well, it is clear that would be better to ignore it if not found, however, I don't see the problem in leaving the warn.

As both options work for me, I think that we should follow your suggestion and remove the warn. Can you do that @kobros-tech ?

kobros-tech:
thank you both, sometimes I lose focus and I admit I just understood now what pedro said about consol log warn.
If I read the comment again I will understand, but when my head is overloading I can be slow to catch up.

I will work on it right now, some few please...

pedrobaeza (approved):
/ocabot merge patch

kobros-tech:
the new module survey_sign_oca is ready for your review and comments, thanks!

@pedrobaeza 
@etobella

PR #103, opened 2025-04-04
[FIX] sign_oca: restricted company logo max-height to prevent l…

anasassi119:
…arge logos from affecting the size of the signing area

### DIFF
--- a/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.xml
+++ b/sign_oca/static/src/components/sign_oca_pdf_portal/sign_oca_pdf_portal.xml
@@ -12,6 +12,7 @@
                                         src="/logo.png"
                                         alt="Logo"
                                         loading="lazy"
+                                        style="height: 70px;"
                                     /></a>
                             </div>
                         </div>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't this still deforming images? What about `max-height`?

anasassi119:
I don't think it would deform images, image deformation occurs when you change the aspect ratio of the image, here, we're simply restricting the height of the logo so that it wouldn't affect the signing area. 

In a signing module, what are your priorities? Even deforming the logo a little bit is fine as long as the signing area is big enough.

An example of the problem this fixes:

**Before**
429812441-8bf7f0a3-93e5-4f6f-9fa6-0fd62638aa44

**After**
429812519-b1ced261-eb3a-451f-a96d-4ff18651a572

pedrobaeza (approved):
OK then

etobella (changes requested):
As stated on https://github.com/OCA/sign/pull/100, use max-height, not height. If you use height with a small logo, the logo will be pixeled.

You will receive the same benefits but other people will not have problems.

Also, fix the commit name (you don't need the [17.0] on the commit, just on the Pull REquest. If possible, avoid two PR, just ammend the current  and push force to replace the current commit.

For example, to fix this you could use:

```shell
git commit --amend
git push -f
```

anasassi119:
Silly me thought I used max-height in this PR. Updated.

etobella (approved):

PR #111, opened 2025-05-24
[18.0][IMP] sign_oca: add kanban/activity view on sign.oca.request.signer

jguenat:
When using the sign systray, the user goes to a kanban view.

- There is no kanban view for this model so Odoo uses a simple incomprehensible one.
- No chatter models inherited on  sign.oca.request.signer so the activity view throws an error.

sign_request_signer

I propose to simply remove the kanban/activity views from the action. Tree and form view is enough IMO.

### DIFF
--- a/sign_oca/views/sign_oca_request.xml
+++ b/sign_oca/views/sign_oca_request.xml
@@ -328,6 +328,7 @@
                 <header>
                     <button
                         string="Sign"
+                        class="btn btn-primary"
                         type="object"
                         name="sign"
                         invisible="not is_allow_signature"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Isn't better to provide both views instead?

jguenat:
Well it could be done but I took a look at previous versions and it wasn't there.
The action views kanaban & activity have been added in the JS during the V17 migration maybe by mistake?

pedrobaeza:
Yeah, maybe, but I think it's worth to add both views.

sign_oca/views/sign_oca_request_signer.xml
```diff
@@ -0,0 +1,131 @@
...
+                <sheet>
+                    <group>
+                        <field name="is_allow_signature" invisible="1" />
+                        <field name="role_id" readonly="1" />
+                        <field name="partner_id" readonly="1" />
+                        <field name="request_id" readonly="1" />
+                        <field name="signed_on" readonly="1" />
+                        <field name="inalterable_hash" />
+                        <field name="latitude" readonly="1" />
+                        <field name="longitude" readonly="1" />
```
jguenat: I added readonly on these fields, makes to sens to edit them in form view right?

sign_oca/views/sign_oca_request_signer.xml
```diff
@@ -0,0 +1,131 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <button
+                        string="Sign"
+                        class="btn btn-primary"
+                        type="object"
+                        name="sign"
+                        invisible="not is_allow_signature"
+                    />
```
jguenat: Added class btn btn-primary for better UX

sign_oca/views/sign_oca_request_signer.xml
```diff
@@ -0,0 +1,131 @@
...
+                        <field name="role_id" readonly="1" />
+                        <field name="partner_id" readonly="1" />
+                        <field name="request_id" readonly="1" />
+                        <field name="signed_on" readonly="1" />
+                        <field name="inalterable_hash" />
+                        <field name="latitude" readonly="1" />
+                        <field name="longitude" readonly="1" />
+                    </group>
+                </sheet>
+                <chatter />
```
jguenat: Not sure about the necessity of a chatter on request.signer, there is already on on the request maybe this will be confusing?

jguenat (commented):
Simple kanban,search and activity view added

some modifications on form view, see below comments

sign_oca/views/sign_oca_request_signer.xml
jguenat: I added it because it what I usually do when addintg activity (& mail.thread) on the model

jguenat:
I created a new file for the sign.oca.request.signer views but that makes the PR confusing to read, let me know if you don't want that or I could do it in two separate commits

pedrobaeza (commented):
I think they can be in the same file as previously, as it's a children model.

sign_oca/views/sign_oca_request.xml
```diff
@@ -339,20 +340,97 @@
                         <field name="role_id" readonly="1" />
                         <field name="partner_id" readonly="1" />
                         <field name="request_id" readonly="1" />
-                        <field name="signed_on" />
+                        <field name="signed_on" readonly="1" />
```
pedrobaeza: Why changing this?

jguenat: it's my understanding that people should sign with the /sign_oca/document route which will take care of the date/latitude/longitude fields.

pedrobaeza: Yeah, but this is unrelated to the PR, isn't it?

jguenat: yeah this originally FIX pull request made to fix small issue when viewing request.signer kind of evolved to IMP pull request adding kanban/activity view so why not also fix the form view?

I can remove it no issue for me

pedrobaeza: Well, at least it should be in a separate commit. What do you think, @etobella ?

etobella: I will do it in another PR. on the migration, a lot of readonly where removed without sense by @peluko00 

https://github.com/OCA/sign/commit/bd3c52d992512785fb54e7dee30407a31af76e4d#diff-4a09fca0e1adda50e9bd55b26df637e63fe1ad2bb513aa2f6f2877df7eb9cbcfL354

It is something out of the scope of this PR.

pedrobaeza: It's not without sense, but moved the readonly to views, as if not, you can't fill that fields through Odoo Excel import mechanism.

jguenat: I removed the readonly change, kept the button primary on form view, renamed commit (could rename PR also)

pedrobaeza (approved):

victoralmau (approved):
Functional review OK

PR #113, opened 2025-05-27
[FIX] sign_oca: Restore readonly fields

etobella:
As commented in #111


pedrobaeza (commented):
Isn't better to put readonly at view level for allowing to import records with that fields (if it's ever a case)?

PR #119, opened 2025-08-25
[16.0][FIX] sign_oca: list requests from recent sent to oldest signed

kobros-tech:
from #85


etobella (commented):
LGTM. Can you make a migration script to add this field? Just to ensure that it doesn't takes too much time to update in a big environment.

kobros-tech:
> LGTM. Can you make a migration script to add this field? Just to ensure that it doesn't takes too much time to update in a big environment.

yes sure, @etobella

kobros-tech:
@etobella 

I have added and tested the migration script for the new field.

sign_oca/migrations/16.0.4.0.2/pre-migration.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Kencove - Mohamed Alkobrosli  <https://kencove.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.add_fields(
```
etobella: Still missing the setting of the field by SQL

pedrobaeza: @etobella this openupgradelib method does the "alter table add column" for you.

pedrobaeza: Oh, you want to say the values... Yes, that's missing, but see my next comment.

pedrobaeza (changes requested):
Instead of creating a sequence field, change `state` values (not state labels) to have a number to "order" them, and order by that field. You should make a migration script for mapping values.

etobella:
> Instead of creating a sequence field, change `state` values (not state labels) to have a number to "order" them, and order by that field. You should make a migration script for mapping values.

Yes, that might be cleaner. Also, we need to set a migration script in any case :wink:

kobros-tech:
> Instead of creating a sequence field, change `state` values (not state labels) to have a number to "order" them, and order by that field. You should make a migration script for mapping values.

ok it is easy to do so, I just ignored as I thought it is not important but the values will be mapped in a next commit

I am trying to get familiar with migration scripts as it could be me next plan, thanks!

kobros-tech:
2025-08-26 17:38:17,579 930498 WARNING sign_oca_16_3 pre-migration: Will run migration script for sign.oca.request 
2025-08-26 17:38:17,580 930498 WARNING sign_oca_16_3 OpenUpgrade: add_fields: There's already an entry for sequence in sign.oca.request. This may mean that there's some misconfiguration, or simply that another module added the same field previously. 
2025-08-26 17:38:17,584 930498 DEBUG sign_oca_16_3 OpenUpgrade: 12 rows affected after 0:00:00.004051 running 
        UPDATE sign_oca_request r
        SET sequence = (
            SELECT ROW_NUMBER() OVER (ORDER BY id)
        )
         
2025-08-26 17:38:17,585 930498 WARNING sign_oca_16_3 pre-migration: <odoo.api.Environment object at 0x76d6c5fd85d0> 


@etobella 
@pedrobaeza

pedrobaeza:
But this is not what we finally asked: https://github.com/OCA/sign/pull/119#pullrequestreview-3154153942

kobros-tech:
@pedrobaeza 

I am really sorry, I am trying hard to understand but I can't

I only know the state field but I don't know which state do you mean?

pedrobaeza:
I mean that field. You have to rename the values of that field to be sequential according what you want to order by. Similar to https://github.com/odoo/odoo/blob/18.0/addons/product/models/product_pricelist_item.py#L46

kobros-tech:
> I mean that field. You have to rename the values of that field to be sequential according what you want to order by. Similar to https://github.com/odoo/odoo/blob/18.0/addons/product/models/product_pricelist_item.py#L46

ah all right, good idea also to apply

thanks Pedro

kobros-tech:
2025-08-27 00:20:07,456 977489 WARNING sign_oca_16_3 OpenUpgrade: add_fields: There's already an entry for state_sequence in sign.oca.request. This may mean that there's some misconfiguration, or simply that another module added the same field previously. 
2025-08-27 00:20:07,457 977489 DEBUG sign_oca_16_3 OpenUpgrade: 12 rows affected after 0:00:00.000827 running 
        UPDATE sign_oca_request
        SET state_sequence = CASE state
            WHEN 'sent' THEN '0_sent'
            WHEN 'draft' THEN '1_draft'
            WHEN 'signed' THEN '2_signed'
            ELSE '3_cancel'
        END
         
2025-08-27 00:20:07,528 977489 INFO sign_oca_16_3 odoo.modules.registry: module sign_oca: creating or updating database tables

kobros-tech:
@pedrobaeza 

I changed the field and updated the code and migration script, functional test was fine to me locally.

pedrobaeza (changes requested):
Don't create a new field, but use the existing one `state`, changing its values (and modifying code accordingly).

kobros-tech:
2025-08-27 18:04:06,006 1024185 DEBUG sign_oca_16_3 OpenUpgrade: 14 rows affected after 0:00:00.001961 running 
        UPDATE sign_oca_request
        SET state = CASE
            WHEN state = 'sent' THEN '0_sent'
            WHEN state = 'draft' THEN '1_draft'
            WHEN state = 'signed' THEN '2_signed'
            WHEN state = 'cancel' THEN '3_cancel'
            ELSE state
        END

kobros-tech:
@pedrobaeza 

Are you sure about last commit I made?

sign_oca/models/sign_oca_request.py
```diff
@@ -25,19 +25,20 @@ class SignOcaRequest(models.Model):
     _name = "sign.oca.request"
     _inherit = ["mail.thread", "mail.activity.mixin"]
     _description = "Sign Request"
+    _order = "state, create_date desc, id desc"
```
pedrobaeza: The other field doesn't add value

```suggestion
    _order = "state, id desc"
```

pedrobaeza (approved):

kobros-tech:
@pedrobaeza 

if you are comfortable I will squash?

pedrobaeza:
OK, go ahead.

kobros-tech:
@pedrobaeza 

I squashed

pedrobaeza (approved):

kobros-tech:
@etobella 

Can you approve this change too, we have 1 approval.
Now, no new fields are needed after using state field and as a result the migration file is not needed anymore.

Thanks

etobella:
One silly question, if we cancel a request, it will be in the top. is this expected?

kobros-tech:
> One silly question, if we cancel a request, it will be in the top. is this expected?

it is not expected at all, I tested the buttons but I will recheck.

Not silly at all, thanks

kobros-tech:
image

kobros-tech:
https://github.com/user-attachments/assets/b3cf1f31-10ae-4f11-94e1-bc21946ae388

kobros-tech:
@etobella 

The behaviour is as intended, canceled requests are the last ones in the list view.

etobella (approved):
LGTM

pedrobaeza:
Please forward port it to upper versions

etobella:
I saw that you ported  it to 18, can you do 17 too please :pray:

risto42:
whoa, this is a seriously big change  for the values of state... I don't believe it's justified, it causes updates in any other module using it not to mention i18n... I don't see any precedent for this type of action either... Please revert and consider a less invasive approach..

kobros-tech:
> whoa, this is a seriously big change for the values of state... I don't believe it's justified, it causes updates in any other module using it not to mention i18n... I don't see any precedent for this type of action either... Please revert and consider a less invasive approach..

can you send the links for the modules of OCA that depend and are affected by this change?

risto42:
I would prefer to see a veritable precedent for adopting this type of 'state' change, which I consider highly unusual.
I have *never* come across such before, not in over 40 years!

There are alternative ways to do this... for example, an very intuitive way (also i18n neutral) would be to create tabs for each state with the last tab being 'all' ... in these cases, filtering is naturally available to assist the user with his workflow.

cheers

PR #124, opened 2025-09-01
[16.0][IMP] sign_oca: add back to draft action button

kobros-tech:
In a competition with: https://github.com/OCA/sign/pull/89

Only one is planned to be merged or we can take the best of both in one single thing.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -254,7 +254,7 @@ class SignOcaRequest(models.Model):
 
     def action_send(self, sign_now=False, message=""):
         self.ensure_one()
-        if self.state != "1_draft":
+        if self.state not in ["1_draft", "0_sent"]:
             return
         self._set_action_log("validate")
         self.state = "0_sent"
@@ -268,14 +268,16 @@ class SignOcaRequest(models.Model):
                 engine="ir.qweb",
                 minimal_qcontext=True,
             )
-            self.env["mail.thread"].message_notify(
-                body=render_result,
-                partner_ids=signer.partner_id.ids,
-                subject=_("New document to sign"),
-                subtype_id=self.env.ref("mail.mt_comment").id,
-                mail_auto_delete=False,
-                email_layout_xmlid="mail.mail_notification_light",
-            )
+            # send the message to partners who didn't sign only
+            if not signer.signed_on:
+                self.env["mail.thread"].message_notify(
+                    body=render_result,
+                    partner_ids=signer.partner_id.ids,
+                    subject=_("New document to sign"),
+                    subtype_id=self.env.ref("mail.mt_comment").id,
+                    mail_auto_delete=False,
+                    email_layout_xmlid="mail.mail_notification_light",
+                )
 
     def action_send_signed_request(self):
         self.ensure_one()
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -235,6 +235,10 @@ class SignOcaRequest(models.Model):
         self.write({"state": "3_cancel"})
         self._set_action_log("cancel")
 
+    def action_reset_to_draft(self):
+        self.write({"state": "1_draft"})
+        self._set_action_log("to_draft")
+
     @api.depends("signer_ids")
     def _compute_signer_count(self):
         for record in self:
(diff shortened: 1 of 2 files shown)


kobros-tech:
https://github.com/user-attachments/assets/0b76b07f-b305-4e46-9195-36c4e5688c92

kobros-tech:
the good in this approach the simple code

the bad is that all signers will be emails even if some of them have already signed, this is bad because they may unsubscribe.

but your opinion is welcome:

@pedrobaeza 
@etobella 
@victoralmau

kobros-tech:
@wlin-kencove

needing to review this one and all merging and their similarities in 18.0

victoralmau (approved):

sign_oca/models/sign_oca_request.py
```diff
@@ -235,6 +235,10 @@ def cancel(self):
         self.write({"state": "3_cancel"})
         self._set_action_log("cancel")
 
+    def action_reset_to_draft(self):
+        self.write({"state": "1_draft"})
+        self._set_action_log("to_draft")
```
etobella, line 240: What will happen if we have 2 signers and one of them has already signed?

kobros-tech: @etobella 

In fact that module can not help unless I cancel and then set it back to draft and then I set to sent, 

For example if a customer and employee, and employee has signed then this workfow will make the email to be sent to the customer or all customers who are in the sign request even if one of them has already signed.

In fact this is the point I made the other PR for, that is making email to be sent only to who did not sign.

kobros-tech:
@etobella 

now this issue is solved after importing a function from my former PR, its in the 2nd commit now.

sign_oca/models/sign_oca_request.py
etobella: But then, all signers should be reseted back, isn't it?

kobros-tech: not at all, the point was just sending the email to who signed will not make sense because if the open the link they find themselves already signed.

but their signature is there not changed, other signers if they sign that will be append to former ones.

To avoid unnecessary emails I select the people who didn't sign to receive the notification email.

But I confirm no partner have to sign a document they already signed as this will make them upset and the purpose of this approach to facilitate to them not to annoy them.

etobella: But that is mainly the issue...

Let me explain myself. If we cancel the sign.request, all signatures should be invalidated and we might need to start over again and sign everything again. 

At least this is the way I see it. For me, the main goal here is to resend the request to the missing signers, wouldn't be better to do that?

kobros-tech: the same approach in the other PR

I just compare this current one to the former one and select the best approach.

I guess I tried to cancel and then rest it and the email is sent to signers who didn't sign and they just sign and the document became fully signed without asking any partner to sign again.

Maybe you can try and make sure if what I assume here is also valid.

I will not decide if this PR is the best or the former one, you reviewers will decide with me as both PRs are ok for me in fact.

victoralmau: IMO, I think another simpler approach would simplify everything: add an icon (if not signed by that signer) to forward in the list of signers

ejemplo

kobros-tech: @etobella 
@victoralmau 
and not change or update the state between canel, draft, and sent?

victoralmau: IMO, no, the status should not be changed. As with other sections of Odoo, why shouldn't it be possible to resend an email to someone without having to change anything?

A button (only visible if sent and not signed) that does just that (sending an email) is the simplest and clearest interface.

kobros-tech: in fact that first PR is doing the same functionality https://github.com/OCA/sign/pull/89

Can we transition to and continue conversation as that button you suggest is there and I don't change the state there too.

victoralmau: Answered at https://github.com/OCA/sign/pull/89, the approach I propose is neither this PR nor the other one.

kobros-tech: I will make it closer there, don't worry.

PR #123, opened 2025-09-01
[FIX] sign_oca: list requests from recent sent to oldest signed

kobros-tech:
Add pre_migration script to be sure about the field update

from:
- https://github.com/OCA/sign/pull/119

PR #125, opened 2025-09-02
[17.0][FIX] sign_oca: list requests from recent sent to oldest signed

kobros-tech:
Add pre_migration script to be sure about the field update

from:
https://github.com/OCA/sign/pull/123

### DIFF
--- a/agreement_sign_oca/models/sign_oca_request.py
+++ b/agreement_sign_oca/models/sign_oca_request.py
@@ -33,7 +33,7 @@ class SignOcaRequest(models.Model):
         )
 
         for request in self:
-            if request.state == "signed" and request.agreement_id and request.data:
+            if request.state == "2_signed" and request.agreement_id and request.data:
                 signed_stage_id = (
                     request.env.company.agreement_sign_oca_signed_stage_id.id
                 )
--- a/agreement_sign_oca/tests/test_agreement_sign_oca.py
+++ b/agreement_sign_oca/tests/test_agreement_sign_oca.py
@@ -156,7 +156,7 @@ class TestAgreementSignOca(BaseCommon):
                 "data": dummy_pdf_content,
                 "name": "Signed Agreement A.pdf",
                 "record_ref": f"agreement,{self.agreement_a.id}",
-                "state": "sent",
+                "state": "0_sent",
                 "signer_ids": [
                     (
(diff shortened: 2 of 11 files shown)


etobella (approved):

kobros-tech:
2025-09-02 15:46:37,343 17524 DEBUG sign_oca_17_1 OpenUpgrade: 8 rows affected after 0:00:00.001123 running 
        UPDATE sign_oca_request
        SET state = CASE
            WHEN state = 'sent' THEN '0_sent'
            WHEN state = 'draft' THEN '1_draft'
            WHEN state = 'signed' THEN '2_signed'
            WHEN state = 'cancel' THEN '3_cancel'
            ELSE state
        END

kobros-tech:
@pedrobaeza 
your approval too, and we merge in 17.0

etobella:
Tests are failing @kobros-tech

kobros-tech:
@pedrobaeza 
@etobella 

I would recommend we improve the addons list in OCA project so that dependent modules do not interfer the tests running for their dependent ones like our case here now, I guess this will make workflow more efficient.

kobros-tech:
> Tests are failing @kobros-tech

This will enforce me to update other modules, we agreed on sign_oca only :)

"agreement_sign_oca"

pedrobaeza:
You have to modify `agreement_sign_oca` to adapt to the new state values.

kobros-tech:
@pedrobaeza 

Can I do this in the same commit and PR?

this current one?

etobella:
@kobros-tech actually, tests are done this way for this reason. Now we are ensuring that nothing is broken when we migrate. It is making fail a test. We might need to fix it. That's all...

pedrobaeza:
> Can I do this in the same commit and PR?
> 
> this current one?

Yes, it should be done in the same PR, but it can be a different commit.

kobros-tech:
@pedrobaeza 
@etobella 

I would like to update other module in this same PR, maybe different commit.

What do you think?

kobros-tech:
@etobella 
@pedrobaeza 

Wow, any comments?

etobella (approved):

pedrobaeza (approved):
Now only 18 is pending:

/ocabot merge minor

PR #127, opened 2025-09-09
[16.0][ADD] hr_personal_equipment_request_sign_oca: new module

quirino95:
New module to allow to generate manual and automatic signature requests from Personal Equipment Request.
Instead of a "static" template, every time a new Personal Equipment Request is created, PPE receipt document is generated and attached in the signature request.
Also, the PPE receipt report template in this module is overwritten, in order to keep the space for signature always in the same position.

PR #137, opened 2025-10-16
[16.0][ADD] hr_personal_equipment_request_sign_oca: Show S/N in PPE Request Report

quirino95:
Depends on https://github.com/OCA/sign/pull/127

Added possibility to show Serial Numbers in PPEs request report file. A related setting is present (default is `False`).
This also means that Sign request cannot be created on Personal Equipment Request Acceptance, but now it's done only when related pickings are completed.
Furthermore, now in report only "Valid" lines are visible, since showing cancelled lines doesn't make sense.


hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
```diff
@@ -16,6 +16,18 @@ class HrPersonalEquipmentRequest(models.Model):
...
+        help="Show S/N in Personal Equipment Request Report",
+    )
+
+    @api.model
+    def _default_show_sn_report(self):
+        return bool(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("hr_personal_equipment_request_sign_oca.show_sn_report", False)
+        )
```
HekkiMelody: **suggestion**: instead of doing this configuration for the entire installation, we could do this per company. What do you think?

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
```diff
@@ -5,10 +5,13 @@ class TestHrPersonalEquipmentRequestSignOca(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.company_id = cls.env.company
+        # flake8: noqa: B950
         cls.template = cls.env.ref(
             "hr_personal_equipment_request_sign_oca.sign_oca_template_personal_equipment_request_demo"
         )
```
HekkiMelody: **suggestion** (non-blocking): you can avoid disabling flake with the comment by splitting the string on two lines

```py
cls.template = cls.env.ref(
            "hr_personal_equipment_request_sign_oca."
            "sign_oca_template_personal_equipment_request_demo"
        )
```

HekkiMelody (commented):
Code review, just a minor comment, otherwise LGTM

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
quirino95: Very useful, thank you!

hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
quirino95: I agree, updated settings.

hr_personal_equipment_request_sign_oca/models/res_config_settings.py
```diff
@@ -0,0 +1,16 @@
...
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    personal_equipment_request_sign_oca_template_id = fields.Many2one(
+        comodel_name="sign.oca.template",
+        related="company_id.personal_equipment_request_sign_oca_template_id",
+        readonly=False,
+    )
+
+    show_sn_report = fields.Boolean(
```
HekkiMelody: **chore**: sorry to nitpick, but settings and company have A LOT of fields from every possible module, so it's good practice to "namespace" the fields by using appropriately specific names. The likelihood of another entirely unrelated module adding a field with this name in the setting is relatively high.

Maybe `ppe_report_show_sn`?

quirino95: Don't worry, you're absolutely right. Yes, `ppe_report_show_sn` is good for me too, so I've updated code.

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
@OCA/human-resources-maintainers Hello, we've been using this in prod for months. Could you take a look? We can forward port after the merge. Thanks!

anusriNPS:
Depends on #156 which resolves precommit issue seen

hr_personal_equipment_request_sign_oca/__manifest__.py
```diff
@@ -5,7 +5,12 @@
     "license": "AGPL-3",
     "author": "PyTech SRL, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sign",
-    "depends": ["sign_oca", "hr_personal_equipment_request", "hr_employee_ppe"],
+    "depends": [
+        "sign_oca",
+        "hr_personal_equipment_request",
+        "hr_employee_ppe",
+        "hr_personal_equipment_stock",
+    ],
```
SirPyTech: **suggestion**: If/when https://github.com/OCA/sign/pull/127 is merged, I would avoid adding a new dependency on a merged module because it might break existing installations that do not have the new dependency available.
Please consider adding a bridge module or including this dependency from the beginning, since https://github.com/OCA/sign/pull/127 isn't merged yet.

hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
```diff
@@ -41,19 +48,15 @@ def _generate_sign_oca_request(self):
                 request_dict = sign_template._prepare_sign_oca_request_vals_from_record(
                     item
                 )
-                report = self.env["ir.actions.report"]._get_report_from_name(
-                    "hr_personal_equipment_request_sign_oca.ppe_sign_report_template"
-                )
-                pdf = report._render_qweb_pdf(report.report_name, [self.id])[0]
-                request_dict.update({"data": base64.b64encode(pdf)})
+                request_dict.update({"data": self._get_pdf_report()})
```
SirPyTech: **issue**: The PDF report should be generated from the single PPE request in the loop.
```suggestion
                request_dict.update({"data": item._get_pdf_report()})
```

hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
```diff
@@ -41,19 +48,15 @@ def _generate_sign_oca_request(self):
...
-    @api.model_create_multi
-    def create(self, vals_list):
-        res = super().create(vals_list)
-        res._generate_sign_oca_request()
-        return res
+    def check_sign_oca_request(self):
+        if all(line.state in ["valid", "cancelled"] for line in self.line_ids) and any(
+            line.state == "valid" for line in self.line_ids
+        ):
+            self._generate_sign_oca_request()
```
SirPyTech: **issue**: If we aren't managing serial numbers, this sign request should still be generated upon record creation: instead of removing the `create` override, please add a check on `ppe_report_show_sn` if the flow should be different.

If we really want to modify the original flow, this change should go in the commit where such flow is added.

hr_personal_equipment_request_sign_oca/models/stock_picking.py
```diff
@@ -0,0 +1,22 @@
+from odoo import models
+
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    def _action_done(self):
+        is_return = any(x.origin_returned_move_id for x in self.move_ids)
```
SirPyTech: **issue**: `_action_done` can be called on multiple pickings: some of them can be returns and others maybe are not.
The problem is that `is_return` is `True` if at least one of them is a return: please identify which picking is a return and which isn't instead.

hr_personal_equipment_request_sign_oca/models/stock_picking.py
```diff
@@ -0,0 +1,22 @@
...
+        res = super()._action_done()
+        if self.equipment_request_id and not is_return:
+            self.equipment_request_id.check_sign_oca_request()
+
+        return res
+
+    def action_cancel(self):
+        is_return = any(x.origin_returned_move_id for x in self.move_ids)
```
SirPyTech: **issue**: `action_cancel` can be called on multiple pickings: some of them can be return and others maybe are not.
The problem is that `is_return` is `True` if at least one of them is a return: please identify which picking is a return and which isn't instead.

hr_personal_equipment_request_sign_oca/models/stock_picking.py
```diff
@@ -0,0 +1,22 @@
...
+        res = super().action_cancel()
+        if self.equipment_request_id and not is_return:
+            self.equipment_request_id.check_sign_oca_request()
```
SirPyTech: **question**: Why should we generate a sign request when a picking is cancelled?

hr_personal_equipment_request_sign_oca/reports/ppe_sign_report_template.xml
```diff
@@ -51,11 +56,16 @@
                                     <tr
                                         t-foreach="doc.line_ids"
                                         t-as="line"
-                                        t-if="line.is_ppe"
+                                        t-if="line.is_ppe and line.state == 'valid'"
```
SirPyTech: **question**: Why is this needed only when serial numbers are added? Maybe this goes in the 1st commit?

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
```diff
@@ -43,13 +46,92 @@ def setUpClass(cls):
             {"name": "Employee Test 2", "user_id": cls.user_2.id}
         )
 
+        cls.location_employee = cls.env["stock.location"].create(
```
SirPyTech: **suggestion**: This setup is the same as https://github.com/OCA/hr/blob/a2251a13108632bdd9c0c5ad3cb9e1ca60f3116f/hr_personal_equipment_stock/tests/test_hr_personal_equipment_stock.py#L15, please consider adding a `common` tests setup over there and reusing it here.

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
```diff
@@ -5,10 +5,13 @@ class TestHrPersonalEquipmentRequestSignOca(TransactionCase):
...
         super().setUpClass()
-        cls.company_id = cls.env.company
         cls.template = cls.env.ref(
-            "hr_personal_equipment_request_sign_oca.sign_oca_template_personal_equipment_request_demo"
+            "hr_personal_equipment_request_sign_oca."
+            "sign_oca_template_personal_equipment_request_demo"
         )
+        cls.company = cls.env.company
+        cls.company.personal_equipment_request_sign_oca_template_id = cls.template
+        cls.warehouse = cls.env.ref("stock.warehouse0")
```
SirPyTech: **chore**: Most of these changes do not belong here, please move them to the previous commit where this piece of code has been added.

hr_personal_equipment_request_sign_oca/models/res_config_settings.py
```diff
@@ -0,0 +1,16 @@
...
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    personal_equipment_request_sign_oca_template_id = fields.Many2one(
+        comodel_name="sign.oca.template",
+        related="company_id.personal_equipment_request_sign_oca_template_id",
+        readonly=False,
+    )
+
+    ppe_report_show_sn = fields.Boolean(
```
SirPyTech: **suggestion**: Please add some documentation for the new behavior, for example: this new flag might be added in the `CONFIGURATION` fragment.

SirPyTech (changes requested):
Thanks for the PR!
I only reviewed the code and looks like adding the serials' management is editing the original code/flow of the module, please see the comments below.

hr_personal_equipment_request_sign_oca/__manifest__.py
SirPyTech: Moved new dependency to a new module

hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
SirPyTech: Applied suggestion

hr_personal_equipment_request_sign_oca/models/stock_picking.py
SirPyTech: signature request only applied for pickings that are not a return

hr_personal_equipment_request_sign_oca/models/stock_picking.py
SirPyTech: signature request only applied for pickings that are not a return

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
SirPyTech: Added `common` test file

hr_personal_equipment_request_sign_oca/tests/test_hr_personal_equipment_request_sign_oca.py
SirPyTech: Changes splitted

hr_personal_equipment_request_sign_oca/models/res_config_settings.py
SirPyTech: Added documentation

hr_personal_equipment_request_sign_oca/models/hr_personal_equipment_request.py
SirPyTech: Added documentation for behavior change

hr_personal_equipment_request_stock_sign_oca/reports/ppe_sign_report_template.xml
```diff
@@ -0,0 +1,23 @@
...
+    >
+        <!-- Show Serial Number after Product column -->
+        <xpath expr="//table/thead//th[1]" position="after">
+            <th t-if="doc.employee_id.company_id.ppe_report_show_sn">
+                S/N
+            </th>
+        </xpath>
+        <xpath expr="//table/tbody//td[1]" position="after">
+            <td t-if="doc.employee_id.company_id.ppe_report_show_sn">
+                <t t-out="line.move_ids.lot_ids.name" />
```
SirPyTech, line 19: **note**: In the previous review I did not notice that `line.lot_ids` does not exist, so I changed it to `line.move_ids.lot_ids`.

SirPyTech (approved):

hr_personal_equipment_request_stock_sign_oca/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Simone Rubino - PyTech
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Signature for Personal Equipment Request with Stock",
```
anusriNPS, line 5: As new module introduced in this PR, PR title needs to be changed accordingly.

hr_personal_equipment_request_stock_sign_oca/models/hr_personal_equipment_request.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models
+
+
+class HrPersonalEquipmentRequest(models.Model):
+    _inherit = "hr.personal.equipment.request"
+
+    def _needs_sign_request(self):
+        needs_sign_request = super()._needs_sign_request()
+        return needs_sign_request and any(
+            equipment.is_ppe and equipment.state == "valid"
```
anusriNPS, line 10: As `super()`checks for `equipment.is_ppe`, do we still need to check again for `equipment.is_ppe` here?

hr_personal_equipment_request_stock_sign_oca/models/stock_picking.py
```diff
@@ -0,0 +1,30 @@
...
+        return_pickings = self.filtered(
+            lambda picking: any(
+                move.origin_returned_move_id for move in picking.move_ids
+            )
+        )
+        no_return_equipment_requests = (self - return_pickings).equipment_request_id
+        if no_return_equipment_requests:
+            no_return_equipment_requests._generate_sign_oca_request()
+
+    def _action_done(self):
```
anusriNPS, line 22: **Issue:** Create a PPE request with two allocations. Validate them and in pickings validation, only one allocation is validated and a backorder is created. What is the expectation in this scenario?

Currently, sign request is created with two equipments value one with serial number assigned and other without serial number.

When i try to cancel the backorder, again a new sign request with two equipments similar to existing report gets generated. I believe we need to conclude expectation in such cases or we must be aware of such cases and could be corrected later as well.

I think we  need not generate sign request when cancel button is handled from picking. Please let me know your views.

hr_personal_equipment_request_stock_sign_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+Generate the Signature Request when a Picking for a Personal Equipment Request is completed.
+
+Since the picking is confirmed after the request creation, only valid allocations are included in the Signature Request.
```
anusriNPS, line 3: **Observation:** Tried to create two allocations in a PPE request and validated picking with only one equipment with serial number and other equipment is moved to cancel state as no backorders were created. However, reports have both equipments listed in the sign request. So, please let me know if i missing something from my end.

Image

Image

Image

anusriNPS (commented):
PR looks great!! Mentioned few observations which could be corrected if required

hr_personal_equipment_request_stock_sign_oca/__manifest__.py
SirPyTech: I can't update the PR title

hr_personal_equipment_request_stock_sign_oca/models/hr_personal_equipment_request.py
SirPyTech: Yes because `super` returns `True` when any line `is_ppe`, here we are checking for lines that are both `is_ppe` and `valid`.
For example if I removed the `is_ppe` check here and we had two lines:
- `is_ppe` and not `valid` (satisfies `super`)
- not `is_ppe` and `valid` (satisfies the updated check without `is_ppe`)

the method would return `True`, but it shouldn't because there is no line that is both `is_ppe` and `valid`.

hr_personal_equipment_request_stock_sign_oca/models/stock_picking.py
SirPyTech: > I think we need not generate sign request when cancel button is handled from picking. Please let me know your views.

Agreed: I removed the override for `stock.picking.action_cancel`.

> Create a PPE request with two allocations. Validate them and in pickings validation, only one allocation is validated and a backorder is created. What is the expectation in this scenario?
>
>Currently, sign request is created with two equipments value one with serial number assigned and other without serial number.

That is because, as you noticed in https://github.com/OCA/sign/pull/137#discussion_r3709748333 too, all PPE lines are printed.
I edited the report to only show the `valid` lines.

hr_personal_equipment_request_stock_sign_oca/readme/DESCRIPTION.md
SirPyTech: That is because all lines are printed in the report,  I edited the report to only show the `valid` lines.

hr_personal_equipment_request_stock_sign_oca/models/stock_picking.py
anusriNPS: With latest report, only `valid` lines are shown. When backorders are used, reports are generated accordingly. `cancel` picking does not generate any sign request. Looks good now.

hr_personal_equipment_request_stock_sign_oca/readme/DESCRIPTION.md
anusriNPS: Agree. It looks fine now.

hr_personal_equipment_request_stock_sign_oca/__manifest__.py
anusriNPS: Agree, seems edit button on PR title is unavailable. we can ignore it.

anusriNPS (approved):

hr_personal_equipment_request_stock_sign_oca/models/hr_personal_equipment_request.py
anusriNPS: Agree, it would be required to filter `lines which are ppe and valid`.

anusriNPS:
Code and Functional Review: LGTM

PR #147, opened 2025-12-23
[IMP] sign_oca: Add error handling for corrupted PDFs

YoussefEttamimi:
- Use strict=False to handle PDFs with minor format issues
- Add try-catch for PDF read operations with clear error messages
- Add try-catch for PDF write operations with helpful user feedback
- Prevent crashes when processing PDFs with encoding issues

This improvement handles PDFs that contain encoding problems or illegal characters that would normally cause PyPDF2 to fail with PdfReadError, providing a more robust document signing experience.

### DIFF
--- a/sign_oca/models/sign_oca_request.py
+++ b/sign_oca/models/sign_oca_request.py
@@ -445,7 +445,16 @@ class SignOcaRequestSigner(models.Model):
         signatory_data = self.request_id.signatory_data
 
         input_data = BytesIO(b64decode(self.request_id.data))
-        reader = PdfFileReader(input_data)
+        try:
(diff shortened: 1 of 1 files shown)


sign_oca/models/sign_oca_request.py
```diff
@@ -464,7 +473,17 @@ def action_sign(self, items, access_token=False, latitude=False, longitude=False
         for page_number in pages:
             output.addPage(pages[page_number])
         output_stream = BytesIO()
-        output.write(output_stream)
+        try:
+            output.write(output_stream)
+        except Exception as e:  # pragma: no cover
```
etobella, line 478: You should remove the pragma no cover in both of your changes and make a test for it.

PR #154, opened 2026-04-10
[ADD] sign_oca: add date field type with calendar picker

lavacano:
#### Summary

Adds a new `date` field type to the sign_oca module, supporting a native browser date picker for the signing flow.

#### Features

- **Calendar picker** — Uses `<input type="date">` for native browser date selection
- **Auto-fill** — Automatically populates with today's date when the field is focused (if empty)
- **PDF rendering** — Date values render as formatted text (MM/DD/YYYY) in the signed PDF, reusing the existing text rendering logic
- **Tab navigation** — Supports keyboard tab navigation between sign fields (same pattern as text fields)

#### Changes

**Backend (`sign_oca_field.py`)**
- Add `("date", "Date")` to the `field_type` selection field

**Backend (`sign_oca_request.py`)**
- Add `_get_pdf_page_date()` method that delegates to the text renderer

**Data (`data.xml`)**
- Add `sign_field_date` seed record so the Date field is available after module install/update

**Frontend (`date.esm.js`) [NEW]**
- New element module following the existing `text.esm.js` pattern
- Registers into the `sign_oca` registry category
- Stores the locale-formatted date string for PDF rendering and the raw ISO value for the date input

**Frontend (`elements.xml`)**
- Add `sign_oca.sign_iframe_field_date` QWeb template with `<input type="date">`

**Manifest (`__manifest__.py`)**
- Register `date.esm.js` in both `web.assets_backend` and `oca_sign.assets_frontend_sign` asset bundles

#### Testing

1. Install/update `sign_oca`
2. Create a sign template, right-click on a page → "Date" field is available in the context menu
3. Open the signing URL as a signer
4. Click the date field — native browser date picker appears
5. Field auto-fills with today's date on focus
6. Complete signing — date renders correctly in the final PDF


etobella (changes requested):
Validating a report fails

PR #160, opened 2026-08-06
[18.0][FIX] sign_oca: Use odoo pdf libraries

### LINKED ISSUE #152: Should migrate to PyPDF 5.4
We have our Odoo deployed on Debian trixie (with Python 3.13.5), and PyPDF2 doesn't get installed by Odoo's `requirements.txt` anymore:

```
PyPDF2==1.26.0 ; python_version <= '3.10'
PyPDF2==2.12.1 ; python_version > '3.10' and python_version < '3.13' # (Noble and below)
PyPDF==5.4.0 ; python_version >= '3.13' # (Trixie)
```

Now, when installing OCA/sign, it requires PyPDF2 at runtime, but doesn't document it in a `requirements.txt` file.

Perhaps it would make sense to import Odoo's [odoo/tools/pdf/](https://github.com/odoo/odoo/blob/18.0/odoo/tools/pdf/__init__.py) directly, as it does the "depend on any PyPDF alternative" dance already

etobella:
Fixes #152

### DIFF
--- a/project_task_sign_oca/tests/test_project_task_sign_oca.py
+++ b/project_task_sign_oca/tests/test_project_task_sign_oca.py
@@ -1,7 +1,8 @@
 # Copyright 2024 - TODAY, Wesley Oliveira <wesley.oliveira@escodoo.com.br>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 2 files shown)


luisDIXMIT (approved):
Code review, LGTM!

PR #164, opened 2026-08-13
[16.0][ADD] sign_docuseal: DocuSeal electronic signature connector

marcos-mendez:
This module connects Odoo to a [DocuSeal](https://www.docuseal.com/) instance, an open-source and self-hostable electronic signature platform. Odoo sends documents out for signature and follows their whole life cycle, without anyone having to open the DocuSeal back office.

It is running in production on a 16.0 database, where it drives the signature of docking contracts.

### What it does

- Sends a document for signature, from a DocuSeal template or from HTML rendered by Odoo.
- Tracks the request in `docuseal.submission`, with one `docuseal.submitter` line per signer and its own status.
- Receives DocuSeal events on an authenticated webhook and applies them idempotently, since DocuSeal retries.
- Stores the signed PDF and the audit trail as attachments on completion.
- Reconciles by cron: a submission still pending after an hour is checked against the API, so a webhook lost in transit does not strand a request.

### Signers stay inside Odoo

Each signer gets a link to an Odoo page, `/docuseal/sign/<submitter>/<token>`, which embeds the DocuSeal form in an iframe. The token is generated by Odoo, one `uuid4` per signer, and compared in constant time.

Besides keeping the signer on a domain they already trust, this avoids mailing out the DocuSeal slug, which is itself a signing credential.

### Two design points worth reviewing

**No dependency on `sign_oca`.** The two modules solve different problems and can be installed side by side. `sign_oca` signs inside Odoo: it owns the field placement, the signature and the certificate. This module hands the document to an external service that owns all three. Mapping one onto the other would mean inventing an abstraction that fits neither. Happy to revisit if the maintainers see it differently.

**Two helpers need a paid DocuSeal plan.** `create_submission_from_html()` (`POST /api/submissions/html`) and `create_template_from_pdf()` (`POST /api/templates/pdf`) answer 404 on a community instance; the client turns that into a plain user error. They are documented as such on the methods and in the README. Everything else works against a community instance.

The webhook accepts its shared secret either in the `X-Odoo-Webhook-Secret` header (documented as the way to do it) or in the URL path. The path form is kept for instances already configured that way, and the README says why it should be migrated: a secret in a URL ends up in web-server logs, proxy logs and referrers.

### Tests

55 tests, 100% coverage on every code file: the API client with `requests` mocked, the submission state machine, and both controllers under `HttpCase` (secret rejection, malformed payload, each event, the retry signal, and the signing page with a valid, an invalid and an unknown token).

Includes a `pt_BR` translation.


pedrobaeza:
What are the missing native features to require another signing system? Why not implementing them here instead? When you connect 2 things, you have 3 problems...

marcos-mendez:
@pedrobaeza It's not a UX gap, it's a legal one. In Brazil an electronic signature only holds up in court under MP 2.200-2/2001 and Law 14.063/2020 when it carries an ICP-Brasil certificate, a PAdES/CAdES signature embedded in the PDF, and an RFC 3161 timestamp from an accredited TSA, with LTV so it stays verifiable after the certificate expires. sign_oca stores a signature image and a certificate report — that is a valid record of consent, but it is not a cryptographically signed document, and it is not what a Brazilian court or a notary asks for.

Implementing that inside sign_oca means building and keeping accredited a PAdES signer, a TSA client, ICP-Brasil chain validation and LTV — that is a compliance product with its own audit cycle, not a feature. DocuSeal is open source and self-hosted, so this is a microservice we run ourselves, not a SaaS dependency; the "3 problems" part is why the module has an idempotent webhook, a reconciliation cron for lost events, and 55 tests at 100% coverage.

Worth trying before deciding: docker run docuseal/docuseal, install the module, send one contract. Happy to keep it out of sign and move it to a Brazilian repo if you prefer, but the two modules do not overlap.

